PR #2, opened 2018-09-10
update field image for rover ruckus

kellyrm:


rbrott:
I'm going to close this because I actually already made the change and was waiting to push the commit (sorry).

PR #37, opened 2020-12-20
Drag-n-drop layout

NoahBres:
Drag-n-drop layout feature

- Added tailwind
- Added styled-components
  - Small styling tweaks
- Implemented `ConfigurableLayout`
  - Added `RadialFab` button and `RadialFabChild` 
  - Added `ViewPicker`
- Added more Roboto font weights
  - 300/500 weights added. Previously only 400. Caused blurry rendering on bold texts (at least on my device, rendering probably differs from device to device as I'm betting the browsers try to fake bold text when font weights aren't available)
  - Change `AutoFitCanvas` resizing behavior to use `ResizeObserver`
    - Improves responsiveness
- Added new icons
  - `create.svg`, `delete_x.svg`, `delete.svg`, `lock_open.svg`, `lock.svg`, (and more now)
- Add unique ID's to`MultipleCheckbox` inputs so you can also click on the label
- Add clarification labels to `GraphView`
  - It seemed quite common that users not know the top right button in the variable selection view switched to the graphing view. Added text letting the users know
  - Added label alerting user that no variables were selected in graph view
- Undo/redo functionality for the grid
- Pause indicator for graph
  - Also added 'k' keypress to pause the graph. A lot of the times the graph can't steal focus and space bar scrolls down the page. YouTube uses both 'space' and 'k' for pausing/unpausing videos for this same reason.
- Red outline on invalid inputs

(I still don't know why git pulls commits from Sep. 15 on every PR, this branch's commits should start on Dec. 13 3db3cad)


rbrott (commented):
Looks great! 🎉  I'm sure a lot of teams will love this feature.

I poked around with it locally, and I just have a few general stylistic/UI suggestions.

* I'm not crazy about the accent divider. 
* The weight-700 typography is a little bold for my tastes; I would prefer 500.
* The body typography is also not consistent. Compare the op mode view and graph view text colors. I personally think the black is easier to read, but I'm willing to compromise.
* The graph button has a nice rounded hover outline that I'd love to see on the other icon buttons.
* Overall, I think the copy should have a slightly more formal tone and direct phrasing.
* Bug: graph view "No variables selected to graph" text is not centered.
* FAB controls: I think pressing the FAB initially should unlock the layout and turn into an "X" that locks the layout when pressed. When unlocked, I don't think the FAB should disappear. In place of the unlock/lock button I think an explicit clear/remove all button would be helpful. 
* Bug: layout add options can go off the screen if the display height is small. Maybe it should be paginated?
* Grid: Remove all but the bottom right handle, add `user-select: none;` when unlocked, and darken the grid lines that the views snap to.
* I'm curious if people would like scrolling layouts. My instinct is no (and the app should truncate views that extend below), but perhaps people will small screens would appreciate it.
* Nit: I like the margin changes between locked/unlocked, but it would be nice for the views to maintain the same overall size. In a sense, they stay in the same position and expand/retract.

Great job! Hopefully I'm not being too picky 😄

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
```diff
@@ -0,0 +1,448 @@
...
+        };
+        return i;
+      }),
+    );
+  };
+
+  const addItem = (item: ConfigurableView) => {
+    // This is set at 6 right now because all the breakpoints are set to 6 columns
+    // Make this dynamic if responsive column breakpoints are set
+    const COLS = 6;
```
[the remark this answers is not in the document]
rbrott: On second thought, I'm happy just removing those breakpoints altogether. I think I prefer that honestly.

FtcDashboard/dash/src/index.jsx
```diff
@@ -1,4 +1,3 @@
-import React from 'react';
```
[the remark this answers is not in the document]
NoahBres: React 17 includes a new JSX babel transform which utilizes custom JSX transform functions instead of translating them to `React.createElement()` so React doesn't need to be in scope for JSX anymore. Apparently it allows them to code split better and have better control over the JSX translation . It's pretty neat.
https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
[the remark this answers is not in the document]
NoahBres: The drag and drop grid library (https://github.com/STRML/react-grid-layout) I am utilizing requires the breakpoint prop for the `ResponsizeGrid` component (which makes it responsive only in width, not height, I think I saw a comment below that I'll need to touch on regarding this). It's nice in theory, at least for most applications. But its attempts to recalculate the layouts when the number of columns change aren't very nice. It's definitely a difficult problem I understand. But because of that it isn't a very smooth experience for a dashboard like application. I would understand the need for having responsive breakpoints if Dashboard was utilized on mobile but that does not seem like the intended audience. 
But anyways, point is that the `ResponsizeGrid` component provides a default columns props that isn't great because of the awkward grid resizing (in my opinion. you can just pass in your own column prop to test it out) so the columns breakpoints are required to manually keep it at 6.

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
```diff
@@ -0,0 +1,448 @@
...
+      const maxX = Math.max(
+        ...gridItems
+          .filter((e) => e.layout.y + e.layout.h === gridMaxY)
+          .map((e) => e.layout.x + e.layout.w),
+      );
+      if (maxX <= COLS - ITEM_WIDTH) {
+        desiredX = maxX;
+        gridMaxY -= ITEM_HEIGHT;
+      }
+    }
```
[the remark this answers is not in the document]
NoahBres: I'm probably missing the entire point but I'm not quite sure what the logic you posted is supposed to do.
The original logic is supposed to insert the item into the lowest row the user has if it has enough space for an item.

Screen Shot 2020-12-21 at 2 36 54 AM

Screen Shot 2020-12-21 at 2 40 05 AM

Screen Shot 2020-12-21 at 2 41 01 AM

Screen Shot 2020-12-21 at 2 43 53 AM

I'm not familiar with the `Math.max([array of numbers])` syntax. It doesn't seem to make JS happy when I copy-paste your code. Is it supposed to find the bottom-right-most element? If so, I'm not sure how you go about doing that without accounting for the height of the elements.

FtcDashboard/dash/src/containers/Dashboard.jsx
```diff
@@ -74,7 +74,11 @@ class Dashboard extends Component {
 Dashboard.propTypes = {
   isConnected: PropTypes.bool.isRequired,
   pingTime: PropTypes.number.isRequired,
-  layoutPreset: PropTypes.oneOf(Object.keys(LayoutPreset)).isRequired,
+
+  // This should be
+  // PropTypes.oneOf(Object.keys(LayoutPreset)).isRequired
+  // but for some reason it breaks
+  layoutPreset: PropTypes.any.isRequired,
```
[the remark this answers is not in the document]
NoahBres: That specific one was an error on my part actually. That change has been reverted.
In earlier commits I was reading `layoutPreset` in the individual views through the middleware doing so would not allow me to use the prop type commented out. I'm not super familiar with redux so I believe it's some dispatching difference? It seems I must have copy pasted the prop types from the views into the `Dashboard` object at one point.

NoahBres:
> I'm not crazy about the accent divider.

Oops. That wasn't supposed to leave my local branch.

> add user-select: none; when unlocked, and darken the grid lines that the views snap to.

What's the `user-select: none` for? 
What do you mean by grid lines? There aren't any as far as I know? Do you mean the dot grid? Because that's actually totally random. It's just a dot grid pattern on the background div so it's totally arbitrary. I can try and dynamically resize them so that it matches up with the actual grid.

> I'm curious if people would like scrolling layouts. My instinct is no (and the app should truncate views that extend below), but perhaps people will small screens would appreciate it.

I actually don't think I can prevent that. The grid library I am using for it (https://github.com/STRML/react-grid-layout) doesn't have any options to constrain the rows. I also wanted to limit the scrolling but without digging into the library internals or just recreating drag and drop functionality from scratch I can't achieve the effect that I want.

NoahBres:
https://user-images.githubusercontent.com/6739076/102788147-cc1bd280-4367-11eb-9d49-d093fd60eba4.mp4

rbrott:
> What's the `user-select: none` for?

Maybe I'm clumsy but I often accidentally select text sometimes instead of dragging the view. It's just a minor annoyance for me.

> What do you mean by grid lines? There aren't any as far as I know? Do you mean the dot grid? Because that's actually totally random. It's just a dot grid pattern on the background div so it's totally arbitrary. I can try and dynamically resize them so that it matches up with the actual grid. And increase contrast on those if that's what you're asking.

Yeah I meant the dot grid. I don't really see the point if it doesn't match the actual grid. I think it would be fine without the dots given the grid's coarseness.

> I actually don't think I can prevent that. The grid library I am using for it (https://github.com/STRML/react-grid-layout) doesn't have any options to constrain the rows. I also wanted to limit the scrolling but without digging into the library internals or just recreating drag and drop functionality from scratch I can't achieve the effect that I want.

OK it's completely fine the way it is then.

rbrott:
> This is the current behavior I have right now.

I like it! In my head, I imagined the FAB states reversed (click pencil to enter edit/unlocked mode, click lock to exit edit mode).

rbrott:
I noticed a few more minor things when driving the UI around. 

* The configuration view seems to be missing some padding that the other views have. Maybe that can go into `<BaseView />`. 
* I also like "Custom" instead of "Configurable" for the dropdown name (code names can stay the same).
* If you toggle unlock/lock quickly you get the following error: 
Screen Shot 2020-12-21 at 9 27 51 AM

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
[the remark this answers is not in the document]
rbrott: Sorry I should have tested more thoroughly. Here's an improved version:
```javascript
    let [newItemBot, newItemLeft] = gridItems
      .map((e) => [e.layout.y + e.layout.h, e.layout.x + e.layout.w])
      .reduce((e, acc) => (e > acc ? e : acc), [ITEM_HEIGHT, 0]);
    if (newItemLeft >= COLS) {
      // final row overflow
      newItemLeft = 0;
      newItemBot += ITEM_HEIGHT;
    }

    setGridItems([
      ...gridItems,
      {
        id: uuidv4(),
        view: item,
        layout: {
          x: newItemLeft,
          y: newItemBot - ITEM_HEIGHT,
          w: ITEM_WIDTH,
          h: ITEM_HEIGHT,
          isDraggable: !isLayoutLocked,
          isResizable: !isLayoutLocked,
        },
      },
    ]);
```

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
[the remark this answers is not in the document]
rbrott: So we can replace the responsive grid with the normal one?

NoahBres:
> Maybe I'm clumsy but I often accidentally select text instead of dragging the view. It's just a minor annoyance for me.

Gotcha. 

>  I think it would be fine without the dots given the grid's coarseness.

The dot grid is there just because I think it looks nice :P But I do like it being utilitarian. Gonna try to resize to match up with the actual grid.

> I like it! In my head, I imagined the FAB states reversed (click pencil to enter edit/unlocked mode, click lock to exit edit mode).

I've always default to buttons showing the current state but I did some digging and the consensus among UI designers seems to lean towards buttons depicting the next action so gonna flip it 👍 

> The configuration view seems to be missing some padding that the other views have. Maybe that can go into <BaseView />

That was actually intentional. It's so the scroll bars line up with the edge of the container in both the custom layout and the normal layouts. The default padding also padded the scrollbars so it looked odd.

NoahBres:
I can't seem to recreate the error you saw when toggling quickly. Is it related to this?

Screen Shot 2020-12-21 at 4 35 45 PM

There's an error when the camera view undergoes a resize and I've yet to look into that. But it seems to be a different error.

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
```diff
@@ -191,17 +197,151 @@ const DEFAULT_GRID_TALL: GridItemType[] = [
...
+      const newGridStateHistory = [...state.gridStateHistory];
+
+      if (state.currentHistoryPosition !== newGridStateHistory.length - 1) {
+        newGridStateHistory.splice(
+          state.currentHistoryPosition + 1,
+          newGridStateHistory.length - state.currentHistoryPosition,
+        );
+      }
+
+      newGridStateHistory.push(action.payload);
```
rbrott, line 268: Does this work?
```javascript
const newGridStateHistory = [
    ...state.gridStateHistory.slice(0, state.currentHistoryPosition + 1),
    action.payload
];
```

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
```diff
@@ -191,17 +197,151 @@ const DEFAULT_GRID_TALL: GridItemType[] = [
...
+      const newActionHistory = [
+        ...state.actionHistory,
+        StateHistoryCommand.APPEND,
+      ];
+      const newCurrentHistoryPosition = state.currentHistoryPosition + 1;
+      const newCurrentHead = newGridStateHistory[newCurrentHistoryPosition];
```
rbrott, line 274: Should the current head be stored? It seems like it's always `gridStateHistory[currentHistoryPosition]` but maybe I'm missing something

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
```diff
@@ -191,17 +197,151 @@ const DEFAULT_GRID_TALL: GridItemType[] = [
...
+  actionHistory: StateHistoryCommand[];
+  currentHistoryPosition: number;
+  currentHead: GridItemType[];
+}
+
+const stateHistoryReducer = (
+  state: StateHistoryReducerState,
+  action: StateHistoryAction,
+): StateHistoryReducerState => {
+  if (action.type === StateHistoryCommand.APPEND) {
```
rbrott, line 225: Can this become `switch`? Should help with exhaustiveness.

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
```diff
@@ -191,17 +197,151 @@ const DEFAULT_GRID_TALL: GridItemType[] = [
...
+        return true;
+      return JSON.stringify(arr1) === JSON.stringify(arr2);
+    };
+
+    let shouldAppend = false;
+
+    if (state.gridStateHistory.length === 0) {
+      shouldAppend = true;
+    } else {
+      shouldAppend = true;
```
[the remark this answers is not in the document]
rbrott: After reading on, I think I prefer returning prematurely instead of setting `shouldAppend = false` and removing the flag entirely is better.

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
```diff
@@ -191,17 +197,151 @@ const DEFAULT_GRID_TALL: GridItemType[] = [
...
+        gridStateHistory: newGridStateHistory,
+        actionHistory: newActionHistory,
+        currentHistoryPosition: newCurrentHistoryPosition,
+        currentHead: newCurrentHead,
+      };
+    } else {
+      return { ...state };
+    }
+  } else if (action.type === StateHistoryCommand.UNDO) {
+    if (state.currentHistoryPosition > 0) {
```
rbrott, line 286: I slightly prefer returning prematurely when `state.currentHistoryPosition <= 0`.

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
```diff
@@ -191,17 +197,151 @@ const DEFAULT_GRID_TALL: GridItemType[] = [
...
+        gridStateHistory: state.gridStateHistory,
+        actionHistory: newActionHistory,
+        currentHistoryPosition: newCurrentHistoryPosition,
+        currentHead: newCurrentHead,
+      };
+    }
+
+    return { ...state };
+  } else if (action.type === StateHistoryCommand.REDO) {
+    if (state.currentHistoryPosition < state.gridStateHistory.length - 1) {
```
rbrott, line 304: ditto

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
```diff
@@ -191,17 +197,151 @@ const DEFAULT_GRID_TALL: GridItemType[] = [
...
+        actionHistory: newActionHistory,
+        currentHistoryPosition: newCurrentHistoryPosition,
+        currentHead: newCurrentHead,
+      };
+    }
+
+    return { ...state };
+  }
+
+  return { ...state };
```
rbrott, line 323: Any reason for this instead of `return state;`? Mutation is still avoided.

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
```diff
@@ -191,17 +197,151 @@ const DEFAULT_GRID_TALL: GridItemType[] = [
...
+type StateHistoryAction =
+  | {
+      type: StateHistoryCommand.APPEND;
+      payload: GridItemType[];
+    }
+  | { type: StateHistoryCommand.UNDO }
+  | { type: StateHistoryCommand.REDO };
+
+interface StateHistoryReducerState {
+  gridStateHistory: GridItemType[][];
```
rbrott, line 215: I guess some of my confusion with the logic traces back to the state representation. It seems like you can accomplish everything with just `gridStateHistory` and `currentHistoryPosition`. Undo/redo just decrement/increment the position and append clears the post-position tail and inserts the new grid state. Am I missing something?

rbrott:
You don't have to move the commits, but in the future I would prefer 711dfd33c971929632ecfa4de999ba333276ce5e to have their own PR.

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
```diff
@@ -239,15 +379,43 @@ export default function ConfigurableLayout() {
...
     })();
     setIsLayoutLocked(!newGridItems.every((e) => e.layout.isResizable));
-    setGridItems(newGridItems);
+    dispatch({ type: StateHistoryCommand.APPEND, payload: newGridItems });
   }, []);
 
+  useEffect(() => {
+    const keyDownListener = (e: KeyboardEvent) => {
+      if (!isLayoutLocked) {
+        if (window.navigator.userAgent.indexOf('Mac') != 1) {
```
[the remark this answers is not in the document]
NoahAndrews: The normal comparison semantics apply even in TS, so strict equality is usually what you want.

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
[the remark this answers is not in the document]
NoahBres: You're absolutely right I should've realized this :P

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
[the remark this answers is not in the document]
NoahBres: It works really nicely on the first two rows. However it breaks down afterwards.
https://user-images.githubusercontent.com/6739076/102969273-03e75f00-44bb-11eb-860e-201037e21f7c.mp4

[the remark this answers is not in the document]
NoahBres: It seems to break down because of this https://stackoverflow.com/a/16567233/3360147

> A common assumption that arrays are compared element-wise is not true:

Screen Shot 2020-12-23 at 1 40 08 AM
 
 JS is great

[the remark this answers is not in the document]
NoahBres: It works very nicely with a custom array comparison function.


```javascript
    // [number] > [number]
    const moreThanNumberArray = (left: number[], right: number[]): boolean => {
      if (left.length == 0) return false;
      // Follows JS native comparison
      else if (right.length == 0) return true;
      // Follows JS native comparison
      else if (left[0] === right[0] && left.length > 1)
        return moreThanNumberArray(left.slice(1), right.slice(1));
      else return left[0] > right[0];
    };

    let [newItemBot, newItemLeft] = currentHead
      .map((e) => [e.layout.y + e.layout.h, e.layout.x + e.layout.w])
      .reduce((e, acc) => (moreThanNumberArray(e, acc) ? e : acc), [
        ITEM_HEIGHT,
        0,
      ]);

    if (newItemLeft >= GRID_COL) {
      // final row overflow
      newItemLeft = 0;
      newItemBot += ITEM_HEIGHT;
    }
```

The polymorphic comparison is very clean. Is that fine as is or do you have any further improvements?

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
```diff
@@ -191,17 +197,151 @@ const DEFAULT_GRID_TALL: GridItemType[] = [
...
+const stateHistoryReducer = (
+  state: StateHistoryReducerState,
+  action: StateHistoryAction,
+): StateHistoryReducerState => {
+  if (action.type === StateHistoryCommand.APPEND) {
+    const arrEquals = (arr1: GridItemType[], arr2: GridItemType[]) => {
+      if ((arr1.length === 0 && arr2.length === 0) || (!arr1 && !arr2))
+        return true;
+      return JSON.stringify(arr1) === JSON.stringify(arr2);
+    };
```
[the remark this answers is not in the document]
NoahBres: I didn't realize there was a lodash dependency. Sweet.

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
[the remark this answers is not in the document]
NoahBres: If you're referring to my use of double equals in the `keyDownListener`, that was just carelessness. I normally just have the eslint and prettier triple equals rules enabled so never really think about them. It seems I have forgotten to install those rules for this project :P 
My apologies.

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
```diff
@@ -258,12 +426,12 @@ export default function ConfigurableLayout() {
 
     let desiredX = 0;
 
-    let gridMaxY = Math.max(...gridItems.map((e) => e.layout.y + e.layout.h));
```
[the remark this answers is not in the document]
NoahBres: Oops didn't see that.

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
NoahBres: Perfectly!

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
NoahBres: I opted not to use a switch for a specific reason. I quite remember why though.

NoahBres: Ah I found the reason. 
eslint's default rules won't allow declarations in case blocks (https://eslint.org/docs/rules/no-case-declarations). It looks like I case just wrap them in brackets and it should be fine apparently.

NoahBres: I'm changing it to a switch statement as that seems more logical but if I may ask, how does a switch statement help with exhaustiveness? It doesn't seem like JS or TS can do exhaustive checks like Kotlin's when statements.

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
NoahBres: I set `currentHead` earlier because I didn't really want to expose `currentHistoryPosition` but didn't really know how to go about that with a reducer (afaik you can't really have an internal/private state?). But now that it's a hook I can rid of it.

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
NoahBres: None. Just stuck those in there with the intention of setting the values inline but never did and never removed it.

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
NoahBres: Is this in reference to the `currentHead` comment from earlier? If so, I'm just going to copy paste my reply here because I resolved that conversation and I don't think GitHub alerts you when new comments are made in resolved threads. As far as I know? I don't see any UI indicators on my end but I may be wrong.

> I created `currentHead` earlier because I didn't really want to expose `currentHistoryPosition` or `gridStateHistory` because I didn't want one to traverse history themselves. But didn't really know how to go about that with a reducer (afaik you can't really have an internal/private state?). But now that it's a hook I can rid of it.

NoahBres:
> You don't have to move the commits, but in the future I would prefer 711dfd3 to and other unrelated changes to have their own PR.

That was supposed to be in its own branch. I messed up some cherry picking. My apologies. I reverted it anyways because there are two more commits after that fixing two-way syncing for the inputs so that commit being in this branch actually breaks things.

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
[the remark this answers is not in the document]
rbrott: Thanks for fixing the array comparison bug... this is getting embarrassing 😳

There are still (at least) two issues with that code. The first is the last conditional which only works in the normal even case. The second is more subtle and involves the vertical calculation. Consider the following grid.
image
In this case, we cannot match bottoms with the short item and have to push it further down. This also exposes a problem with my original reduce initializer.

Here's the solution I came up with:
```javascript
// find the bottom, right grid item and tentatively place the new item to its right with bottoms aligned
    let [newItemBotMin, newItemLeft] = gridItems
      .map((e) => [e.layout.y + e.layout.h, e.layout.x + e.layout.w])
      .reduce(maxArray, [0, 0]);

    // if this placement puts the new item off the screen, push it to the row below
    if (newItemLeft + ITEM_WIDTH > COLS) {
      newItemLeft = 0;
      newItemBotMin += ITEM_HEIGHT;
    }

    // find the minimum top for the new item to avoid intersecting items above
    const newItemTopMin = gridItems
      .filter((e) =>
        intervalsIntersect(
          [newItemLeft, newItemLeft + ITEM_WIDTH],
          [e.layout.x, e.layout.x + e.layout.w],
        ),
      )
      .map((e) => e.layout.y + e.layout.h)
      .reduce((bottom, acc) => Math.max(bottom, acc), 0);

    // adjust the original new item bottom if necessary given the minimum top
    const newItemTop = Math.max(newItemBotMin - ITEM_HEIGHT, newItemTopMin);

    setGridItems([
      ...gridItems,
      {
        id: uuidv4(),
        view: item,
        layout: {
          x: newItemLeft,
          y: newItemTop,
          w: ITEM_WIDTH,
          h: ITEM_HEIGHT,
          isDraggable: !isLayoutLocked,
          isResizable: !isLayoutLocked,
        },
      },
    ]);
```
And the utility functions are
```javascript
function maxArray(a: number[], b: number[]) {
  if (a.length != b.length) {
    throw new Error('cannot compare arrays with different lengths');
  }
  for (let i = 0; i < a.length; i++) {
    if (a[i] > b[i]) {
      return a;
    } else if (b[i] > a[i]) {
      return b;
    }
  }
  return a;
}

function intervalsIntersect([a, b]: number[], [c, d]: number[]) {
  return Math.max(a, c) < Math.min(b, d);
}
```

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
rbrott: The [docs](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#union-exhaustiveness-checking) aren't super helpful, but I was able to remove the default clause without issue.

FtcDashboard/dash/src/hooks/useUndoHistory.tsx
```diff
@@ -0,0 +1,161 @@
...
+import { isEqual } from 'lodash';
+
+enum StateHistoryCommand {
+  INITIALIZE,
+  APPEND,
+  UNDO,
+  REDO,
+}
+
+export default function useUndoHistory<T>(
```
rbrott, line 11: I think the hook is a big improvement. Good job 👍

FtcDashboard/dash/src/hooks/useUndoHistory.tsx
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  const stateHistoryReducer = (
+    state: StateHistoryReducerState,
+    action: StateHistoryAction,
+  ): StateHistoryReducerState => {
+    switch (action.type) {
+      case StateHistoryCommand.INITIALIZE: {
+        const history = [...state.history];
+        history[0] = action.payload;
```
rbrott, line 47: I don't see why you want the rest of the history. Replace with `const history = [action.payload]`?

FtcDashboard/dash/src/hooks/useUndoHistory.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+        return {
+          ...state,
+          history: history,
```
rbrott, line 51:
```suggestion
          history,
```

FtcDashboard/dash/src/hooks/useUndoHistory.tsx
```diff
@@ -0,0 +1,161 @@
...
+    | {
+        type: StateHistoryCommand.APPEND;
+        payload: T;
+      }
+    | { type: StateHistoryCommand.UNDO }
+    | { type: StateHistoryCommand.REDO };
+
+  interface StateHistoryReducerState {
+    history: T[];
+    actionHistory: StateHistoryCommand[];
```
rbrott, line 36: I'm going to try to explain why I think `actionHistory` is unnecessary.

On init, set `history = [payload], pos = 0`.
On append, set `history = [...history.slice(0, pos + 1), payload], pos = pos + 1`.
On undo, set `pos = max(0, pos - 1)`
On redo, set `pos = min(pos + 1, history.length - 1)`

Maybe I'm missing some subtleties of the current semantics.

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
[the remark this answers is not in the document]
NoahBres: This works really well! Do you want to keep the utility functions in the `addItem()` method or move them to a separate file?

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
NoahBres: Ah yea. Now it's throwing compile errors when I remove default and another case. Neat!

FtcDashboard/dash/src/hooks/useUndoHistory.tsx
NoahBres: My reasoning was that there is nothing really preventing one from dispatching initialize after multiple operations. So calling append then initialize would reset the history. When I wrote that I thought not resetting all of history would be the best behavior but now I'm realizing that makes the most sense as subsequent operations should depend on the preceding event so transparently replacing the initial state is bad.

FtcDashboard/dash/src/hooks/useUndoHistory.tsx
NoahBres: So the reason that history needs to keep track of actions is because of this comparison in the `APPEND` case:

Screen Shot 2020-12-24 at 12 37 41 AM

It's the entire reason why I had to rewrite the entire history functionality using the convoluted-ish reducer. I started out with storing history and current pos in refs and using the logic you just described. Would've been a lot simpler/shorter. It was a big point of frustration having to discover that it didn't work :P

But essentially, the grid state has a two-way data binding with `ReactGridLayout` so `RGL` would set `gridItems` when the `onLayoutChanged` listener is called. But what happens is when you dispatch an `UNDO` or `REDO` command, it'll set the grid items which will then in turn call `onLayoutChanged` which dispatches `APPEND`. So every time `UNDO` or `REDO` was called, it would dispatch a duplicate `APPEND` command, since `APPEND` commands are just fired when `onLayoutChanged` is called. And there's no way to check the source of that command or anything. So you need to filter out these duplicate `APPEND` commands by checking the last command (and checking if they're of equal value).

Maybe there's a much simpler solution to this but I spent a while trying to figure out why undo redo commands wouldn't work with the naive (literally like 8 line) approach. Setting a last command ref doesn't work either because of set state race conditions so the reducer with an action history was ideal.

rbrott: Ok that context is helpful. I think you can just compare the new state with the current state in append and no-op on match?

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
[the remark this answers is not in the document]
rbrott: I like them in the current file at the top level (below imports).

FtcDashboard/dash/src/hooks/useUndoHistory.tsx
NoahBres: I started with that but I no longer remember why that didn't work. I'm 90% sure there was an explicit case it broke on. However, I just commented out the the last command check and left in the equality check and it seems to be working fine. Gonna test more thoroughly but I think I might just end up removing the history if this seems to be fine.

NoahBres: So um...after extensive testing I can no longer replicate the earlier issue that necessitated the action history. It can assure you it was 100% necessary at that point but I'm pretty sure some logic refactoring in one of the past commits probably fixed some faulty logic. Guess I'm removing action history now :P

NoahBres:
Turns out half of the original undo history reducer was simply due to faulty logic and patching over it with poor fixes (although it did work). Refactoring the logic obviated the need for most of them. The end result is so much cleaner and very simple (for the entire PR). Thank you so much for your patience despite the fatuity. I really really appreciate it.

rbrott:
Alright! This is really close. I noticed two things:
* On reload, I think the grid should start out locked.
* Some of the graph text (and maybe canvas) positioning is off for regular layouts. The "Graph" preset is a good example.

This should probably go in a separate PR, but I was thinking about changing some copy to make it more direct and actionable. Here are some possible substitutions:
* "No variables selected to graph" -> "No telemetry selected to graph"
* "Click the button in the upper right to switch to the graphing view" -> "Press the upper-right button to graph selected keys over time"
* "Graph related telemetry packets have not yet been received \ They will appear here once sent" -> "Send number-valued telemetry data to graph them over time"
* "Event loop detached" -> "Op mode controls have not initialized"

rbrott:
Ok I took out the hardware for a final spin. 

* I missed some text: replace "Variables to graph:" with "Telemetry to graph:" (I'm trying to distinguish telemetry from config variables).
* I like the scroll behavior of config, and I was wondering if you could port it to the other scrollable views without too much difficulty.
* This is a super nit but the little arrow in the op mode select dropdown interferes with the names of the longest op modes. 
* The "Press the upper-right button to graph selected keys over time" text should probably move to the top of the graphing view and be a bit bigger (maybe `text-lg`). I guess this applies to the op mode view no initialization text.
* Also it would be great if that op mode text were centered in the container as well.
* Pause indicator is a great feature, but I think it would be better as a button that users can press as well as use the shortcut (click the double bars to pause, click triangle to resume).

I _think_ this is the last batch of requests.

rbrott:
Sigh. I forgot to mention that the node version in `FtcDashboard/build.gradle` needs to be bumped from `12.9.0` to `12.13.0` for the new dependencies.

NoahBres:
> Sigh. I forgot to mention that the node version in FtcDashboard/build.gradle needs to be bumped from 12.9.0 to 12.13.0 for the new dependencies.

I didn't even notice that gradle managed the compilation. That's really neat.

> I like the scroll behavior of config, and I was wondering if you could port it to the other scrollable views without too much difficulty.

Do you mean how the main content scrolls on overflow separately from the heading?

> Pause indicator is a great feature, but I think it would be better as a button that users can press as well as use the shortcut (click the double bars to pause, click triangle to resume).

What are your thoughts on putting the play/pause button there? (Ignore the offset. That's because I used unicode characters instead of looking for an icon as a quick mockup)
Or do you prefer having it float on the graph?

Screen Shot 2020-12-25 at 4 59 19 AM

Happy holidays!

rbrott:
> Do you mean how the main content scrolls on overflow separately from the heading?

Yes, exactly.

> What are your thoughts on putting the play/pause button there? (Ignore the offset. That's because I used unicode characters instead of looking for an icon as a quick mockup)
> Or do you prefer having it float on the graph?

I like it best on the right with the close button, but I don't really mind it beside the "Graph" text. 

Happy holidays as well!

Issue #34, opened 2020-12-10
Expand input validation 


NoahBres:
- Handle scientific notation
- Handle 0.000
- Need to look for more edge cases

rbrott:
Ready to close?

PR #39, opened 2021-01-03
Improved inputs

NoahBres:
Issue #34 

- Change input value to a local state value
- Sync state on prop change
- Implement a more forgiving double validation using `!isNaN(Number(val))` instead of regex
    - Cases like `1.`, `.0`, `0.010` become valid
    - No scientific notation conversion in the middle of typing unless prop is changed explicitly via dashboard server side 
- Red border on invalid inputs
    - Previously only had red focus. Border persists when input isn't in focus
- Fix if else branch fo integer validation (sometimes returned null because some cases wouldnt fit either case)

https://user-images.githubusercontent.com/6739076/103490688-2a729780-4de3-11eb-97ef-c2e0d3d8d638.mov


FtcDashboard/dash/src/components/inputs/validation.js
```diff
@@ -1,34 +1,30 @@
 // TODO: extract patterns into constants?
```
rbrott, line 1: Guess this comment is now irrelevant and can be removed

FtcDashboard/dash/src/components/inputs/validation.js
```diff
@@ -1,34 +1,30 @@
...
-    };
-  } else if (/^-?\d*$/.test(value)) {
+  if (/^-?\d*$/.test(value) && value !== '-' && value !== '') {
     return {
       value: parseInt(value, 10),
       valid: true,
     };
+  } else {
+    return {
+      value: value,
```
rbrott, line 25:
```suggestion
      value,
```

FtcDashboard/dash/src/components/inputs/validation.js
```diff
@@ -1,34 +1,30 @@
...
 };
 
 export const validateInt = (value) => {
-  if (value === '-' || value === '') {
-    return {
-      value: value,
-      valid: false,
-    };
-  } else if (/^-?\d*$/.test(value)) {
+  if (/^-?\d*$/.test(value) && value !== '-' && value !== '') {
```
rbrott, line 18:
```suggestion
  if (/^-?\d+$/.test(value)) {
```

NoahBres: Those are actually necessary because it causes the following error:

Screen Shot 2021-01-03 at 7 49 36 PM

It's because `parseInt('')` or `parseInt('-')` return `NaN` so they should be invalid

rbrott: Oh yeah I was thinking the `+` instead of `*` would incorporate that logic into the regex.

NoahBres: Ah. I missed that part :P

rbrott:
Ready to merge?

NoahBres:
Yep

PR #41, opened 2021-01-04
Improved typings

NoahBres:
Improves typings within dashboard source. Improves future developer experience as one can utilize type completions and confidently know the shape of the data.

- Install redux and redux logger typings
- Move redux related code to a `store` folder
- Convert actions, reducers, and middleware to Typescript
- Create a `types` folder in the `store` folder to hold redux relevant types
- Convert certain views to Typescript
    - Convert `ConfigView`, `GraphView`, `OpModeView`, and `TelemetryView`
    - Not a comprehensive conversion of all the views. These just serve as an initial test to ensure that all the typings work. Future views can be converted as one works on them.

Still have no clue why my PR's include extraneous commits. Relevant commits begin at 4806f7b


FtcDashboard/dash/src/containers/OpModeView.tsx
```diff
@@ -1,28 +1,59 @@
...
+  BaseViewHeadingProps,
 } from './BaseView';
 
 import { ReactComponent as GamepadIcon } from '../assets/icons/gamepad.svg';
 
+type OpModeViewState = {
+  selectedOpMode: string;
+};
+
+// TODO move this elsewhere
```
NoahBres, line 25: Do you have any opinions on this constant? I feel like it should be moved elsewhere. Specifically because I also utilize it in the `LoggingView` concept I sent earlier. So a singular shared location seems more apt as this seems to be part of general OpMode behavior rather than something specific to this view.

FtcDashboard/dash/src/store/middleware/storageMiddleware.ts
```diff
@@ -0,0 +1,30 @@
+import { Middleware } from 'redux';
+
+import LayoutPreset, { LayoutPresetType } from '../../enums/LayoutPreset';
+import { GET_LAYOUT_PRESET } from '../types';
+import { receiveLayoutPreset } from '../actions/settings';
+import { RootState } from '../reducers';
+
+// TODO move this elsewhere
```
NoahBres, line 8: Likewise with this constant? This doesn't really have a valid reason for moving other than that it's an arbitrary constant. But it doesn't seem like it should be in middleware. Although I don't really have much experience with Redux so I'm not quite sure what is appropriate for middleware.

FtcDashboard/dash/src/containers/OpModeView.tsx
NoahBres: Oops. I should've picked the actual declaration to comment on rather than the code comment.

rbrott:
Can you merge the current `master` into this branch? I think that will cleanup the diff.

FtcDashboard/dash/src/containers/OpModeView.tsx
rbrott: Sure you can move it elsewhere for reuse, but for now I think it can stay where it is.

FtcDashboard/dash/src/containers/OpModeView.tsx
```diff
@@ -1,28 +1,59 @@
...
+const mapStateToProps = (state: RootState) => ({
+  available: state.status.available,
+  activeOpMode: state.status.activeOpMode,
+  activeOpModeStatus: state.status.activeOpModeStatus,
+  opModeList: state.status.opModeList,
+  warningMessage: state.status.warningMessage,
+  errorMessage: state.status.errorMessage,
+  gamepad1Connected: state.gamepad.gamepad1Connected,
+  gamepad2Connected: state.gamepad.gamepad2Connected,
+});
```
rbrott, line 37: Does this work?
```
const mapStateToProps = ({ status, gamepad }: RootState) => ({
  ...status,
  ...gamepad,
});
```

FtcDashboard/dash/src/containers/OpModeView.tsx
```diff
@@ -32,7 +63,15 @@ class OpModeView extends React.Component {
     this.onChange = this.onChange.bind(this);
   }
 
-  static getDerivedStateFromProps(props, state) {
+  static propTypes = {
+    isDraggable: PropTypes.bool,
+    isUnlocked: PropTypes.bool,
+  };
```
rbrott, line 69: Is this necessary?

FtcDashboard/dash/src/store/middleware/storageMiddleware.ts
rbrott: Maybe it's nice to have all the storage keys in one place? Though I'm content with its current location.

FtcDashboard/dash/src/components/ConfigurableLayout.tsx
```diff
@@ -89,21 +89,21 @@ const Container = styled.div.attrs<{ isLayoutLocked: boolean }>(
   background-position: ${`calc(0.5rem + ${GRID_DOT_PADDING}px) calc(0.5rem + ${GRID_DOT_PADDING}px - 5px)`};
 `;
 
-interface GridItem {
+type GridItem = {
```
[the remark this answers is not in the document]
NoahBres: Commit comment in a74483b explains the superficial seeming changes.

It's less of a best practices and more so this one person's personal opinions. However, I do think he explains his reasoning well. 
https://fettblog.eu/tidy-typescript-prefer-type-aliases/

FtcDashboard/dash/src/containers/BaseView.tsx
```diff
@@ -1,24 +1,28 @@
...
-})`
+type BaseViewHeadingProps = {
+  isDraggable?: boolean;
+};
+
+const BaseViewHeading = styled.h2.attrs<BaseViewHeadingProps>((props) => ({
+  className: `text-xl w-full px-4 py-2 font-medium ${
+    props.isDraggable ? 'grab-handle' : ''
+  }`,
+}))<BaseViewHeadingProps>``;
```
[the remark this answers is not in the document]
NoahBres: It seems so. I don't really like it at all but it seems like using the attrs builder function doesn't properly pass its type on to the tag function. Removing it causes a compilation error as the actual component no longer possesses those prop types. I am not a fan of using the attrs function just for tailwind classes. You also lose intellisense when using the tailwind intellisense vscode component. I think a better option would be to move everything to the actual tag function using the tailwind [@apply syntax](https://tailwindcss.com/docs/functions-and-directives#apply) (much cleaner for components with a lot of utility classes applied anyways). However, you do need use some babel macro like tailwind.macro or twin.macro when paired with styled-components so I haven't gotten around to doing that yet.

FtcDashboard/dash/src/containers/OpModeView.tsx
NoahBres: Apparently not! That's just the original prop-types enforcement converted to typescript. But it seems that eslint no longer complains about lack of proptypes in a typescript file.

PR #42, opened 2021-01-11
PS4 Dualshock Controller Support

NoahBres:
Adds PS4 Dualshock Controller support to the gamepad


FtcDashboard/dash/src/enums/GamepadType.ts
```diff
@@ -0,0 +1,65 @@
+/*
+The gamepad mappings have been tested with the following devices:
+  - 46d-c216-Logitech Dual Action
+  - 46d-c21d-Xbox 360 Wired Controller
```
rbrott, line 4: Perhaps this isn't very useful anymore?

NoahBres: Not particularly. Although the pid/vid for future reference might be useful for future reference?

FtcDashboard/dash/src/enums/GamepadType.ts
```diff
@@ -0,0 +1,65 @@
...
+    } else if (gamepad.id.search('Xbox 360') !== -1) {
+      return GamepadType.XBOX_360;
+    } else if (
+      // Dualshock Gen 1 & 2
+      (gamepad.id.search('054c') !== -1 &&
+        (gamepad.id.search('09cc') !== -1 ||
+          gamepad.id.search('05c4') !== -1)) ||
+      // Etpack Wired Controller
+      // https://github.com/OpenFTC/OpenRC-Turbo/blob/b75e0f8da4925c077c6c6e5d9d119676ebba2c56/Hardware/src/main/java/com/qualcomm/hardware/sony/SonyGamepadPS4.java#L145
+      (gamepad.id.search('7545') !== -1 && gamepad.id.search('104') !== -1)
```
rbrott, line 42: It would be nice if these were constants; it will help distinguish between VIDs and PIDs.

rbrott (commented):
This stretches the scope of the PR, but I now think it's nice to largely mirror the SDK's VID/PID matching logic for all gamepad types. Exceptions like the Gen 2 Dualshock be noted explicitly.

Totally optional, but I think a special icon for unsupported gamepads in the UI would be helpful.

https://github.com/OpenFTC/Extracted-RC

NoahBres:
> This stretches the scope of the PR, but I now think it's nice to largely mirror the SDK's VID/PID matching logic for all gamepad types. Exceptions like the Gen 2 Dualshock be noted explicitly.

Do you mean just replacing the pattern matches for the logitech dual action and xbox 360 with PID/VID searches?

NoahBres:
> Totally optional, but I think a special icon for unsupported gamepads in the UI would be helpful.

You're referring to the opmode view's icons right? There's no way to tell if an unsupported gamepad is connected using the current redux middleware without further changes I believe? Since it just uses the `gamepad1/2Connected` boolean in the middleware state. 

As an alternative, I thought it would be neat to implement toast notifications that appear in the bottom left for certain actions. Initially, I thought it would be useful for undo/redo and new gamepad connections. I think an "Unsupported gamepad" would be a nice addition to that.

rbrott:
> What do you have in mind?
> Do you mean just replacing the pattern matches for the logitech dual action and xbox 360 with PID/VID searches?

Yes, exactly.

> You're referring to the opmode view's icons right? There's no way to tell if an unsupported gamepad is connected using the current redux middleware without further changes I believe? Since it just uses the `gamepad1/2Connected` boolean in the middleware state.

Yes, it requires some store changes.

> As an alternative, I thought it would be neat to implement toast notifications that appear in the bottom left for certain actions. Initially, I thought it would be useful for undo/redo and new gamepad connections. I think an "Unsupported gamepad" would be a nice addition to that.

👍

NoahBres:
> > What do you have in mind?
Do you mean just replacing the pattern matches for the logitech dual action and xbox 360 with PID/VID searches?

> Yes, exactly.

Do you have access to either the xbox 360 or the logitech controller?
I currently don't have access to either (I may have access to the logitech at some point in the future but can't guarantee a time soon) so I would rather not push changes without testing.

rbrott:
> Do you have access to either the xbox 360 or the logitech controller?

No, unfortunately.

PR #47, opened 2021-02-02
fix xbox controller buttons

smartynenko:
Current Xbox 360 button assignment appear to be incorrect and do not match controller return values. Could you update mapping to match [standard gamepad layot](https://w3c.github.io/gamepad/#remapping) ?

### DIFF
--- a/FtcDashboard/dash/src/enums/GamepadType.ts
+++ b/FtcDashboard/dash/src/enums/GamepadType.ts
@@ -45,7 +45,7 @@ export default {
     } else if (
       gamepad.id.search(
         new RegExp(`(?=.*${ETPACK_VID})(?=.*${ETPACK_PID})`, 'i'),
-      )
+      ) !== -1 
     ) {
       return GamepadType.SONY_DUALSHOCK_4;
     } else {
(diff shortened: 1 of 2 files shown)


rbrott:
cc @NoahBres

NoahBres:
I never tested the Xbox 360 controller because of the inability to refactor the id system using PID/VID. I just left the original implementation in. I have no idea whether it follows the standard w3 spec or not. Perhaps the browser implementation may have changed since you originally created it. The spec isn't very solid so that would not be so far-fetched. 

Would you like me to test it? I can get access to one by next week if so. But I'm assuming smartynenko has done so and this PR is based on their own testing.

rbrott:
@smartynenko Can you include the OS, browser, and gamepad ID tuples that you've tested? A comment in the source would be great.

@NoahBres More testing would help, but I'm fine being lazy and waiting for people to report broken mappings.

rbrott:
@smartynenko Thanks. Sorry to be a pain, but I want to be precise. Is the `id` field value `"generic controller (VID=24C6, PID=530A)"`?

smartynenko:
np. are you referring to gamepad.id field? My controller reports identifies itself as `Xbox 360 Controller (XInput STANDARD GAMEPAD)`

rbrott:
Yes. In both browsers? I guess the VID/PID isn't exposed through the API. Also is `gamepad.mapping` `"standard"`?

smartynenko:
correct, same `id=Xbox 360 Controller (XInput STANDARD GAMEPAD)` and `mapping=standard` values in both browsers.

rbrott:
Awesome. It'd be great if you could slip those strings into the comment.

rbrott:
Perfect. Thanks!

PR #55, opened 2021-03-14
Live log

NoahBres:
Live Logging View:
- Grid of telemetry data logged over time
- Resizable columns. Show/hide columns
- Virtual grid using [react-window](https://github.com/bvaughn/react-window)
  - Smooth performance regardless of log size. My laptop can handle 30k items without framerate drop. Didn't bother to test further but it should be able to handle even the most aggressive FTC level data aggregation
- Auto-scrolling
- CSV export
  - includes `telemetry.log()` calls

Misc:
- Extract `Tooltip` component from `RadialFabChild` into a reusable component
- Add `LoggingView` to `ViewPicker`
- Slight visual tweaks to button focus

https://user-images.githubusercontent.com/6739076/111058770-471bf300-8456-11eb-810b-5af49017d860.mp4


FtcDashboard/dash/src/components/ConfigurableLayout.tsx
```diff
@@ -303,17 +305,21 @@ export default function ConfigurableLayout() {
...
-          );
+    const containerResizerObserver = new ResizeObserver(
+      (entries: ResizeObserverEntry[]) => {
+        if (gridWrapperRef.current) {
+          for (const entry of entries) {
+            if (entry.target === gridWrapperRef.current) {
+              const width =
+                gridWrapperRef.current.clientWidth - 2 * GRID_DOT_PADDING;
+              setGridBgSize(
+                width / ((GRID_COL / 4) * Math.floor(width / 300) + 1) / 3,
```
rbrott, line 316: I know this slipped through a previous review, but could you explain what this computation does?

NoahBres: To be completely honest, I should have created a comment on what it does because I don't quite remember. The numbers are arbitrary and are chosen because they looked nice and it's been a while since I wrote that. 
It is meant to scale the grid background size but in increments so that the grid background smoothly scales and lines up with the edges of each view. The `Math.floor(width /300)` is dividing it into integer sections. Then it multiplies it by the column count divided by four (I think divided by four just because it looked nice). I could not explain to you the function of `+ 1` and `/ 3`. The entire thing was just arbitrarily tweaked until it looked pleasing.

PR #63, opened 2021-07-15
Fix dashboard toggle OpMode not checking for stop

dansman805:
If the user stops the Dashboard Enable/Disable OpMode while waitForStart() is running, the current thread gets interrupted. However, interrupting the thread does not halt thread execution, meaning that even if the user hits "stop" during the initialization phase, dashboard's enable state will be toggled. By checking that isStopRequested() is false, this change makes sure that stop has been pressed before toggling dashboard's enable state.

### DIFF
--- a/FtcDashboard/src/main/java/com/acmerobotics/dashboard/FtcDashboard.java
+++ b/FtcDashboard/src/main/java/com/acmerobotics/dashboard/FtcDashboard.java
@@ -667,6 +667,19 @@ public class FtcDashboard implements OpModeManagerImpl.Notifications {
 
                         waitForStart();
 
+                        // If the user stops the OpMode while waitForStart() is
+                        // running, the current thread gets interrupted.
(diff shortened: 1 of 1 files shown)


rbrott (commented):
Thanks for the fix. Sorry to nit, but I prefer the check to return early like [here](https://github.com/acmerobotics/relic-recovery/blob/4ff05bbf906829aef0a98bc32691e5d0eadc1d8f/TeamCode/src/main/java/com/acmerobotics/relicrecovery/opmodes/AutoOpMode.java#L61-L63).

dansman805:
> Thanks for the fix. Sorry to nit, but I prefer the check to return early like [here](https://github.com/acmerobotics/relic-recovery/blob/4ff05bbf906829aef0a98bc32691e5d0eadc1d8f/TeamCode/src/main/java/com/acmerobotics/relicrecovery/opmodes/AutoOpMode.java#L61-L63).

Changed

PR #65, opened 2021-09-21
Updates to SDK 7.0 and fixes crash

serivesmejia:
Using the newly released SDK 7.0, the application crashes when i open ftc dashboard on the browser.

Further investigation of the issue led me to the RobotStatus class which uses the following method that was changed in the new version of the SDK.

image

After adding a .message after the method call in RobotStatus, the error seems to be fixed. This makes the next version of ftc dash not compatible with sdks lower than 7.0, which hopefully shouldn't be an issue. I will try to open another PR to the road runner quickstart ~~today~~ once this is released too.

### DIFF
--- a/FtcDashboard/build.gradle
+++ b/FtcDashboard/build.gradle
@@ -90,10 +90,10 @@ repositories {
 }
 
 dependencies {
-    api 'org.firstinspires.ftc:RobotCore:6.2.1'
-    api 'org.firstinspires.ftc:Hardware:6.2.1'
-    api 'org.firstinspires.ftc:RobotServer:6.2.1'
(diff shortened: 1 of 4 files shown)


NoahAndrews (changes requested):
This is not a complete update to 7.0. Look at how many files were updated on the official repo. Even if you ignore the samples, you're missing changes to several files.
image

serivesmejia:
yeah i was just looking at that but had to go out somewhere. i will try to open another PR appropiately updating the SDK

rbrott:
Thanks @serivesmejia ! I usually like to do SDK updates myself though. I've been a bit busy these past few days, but I'm planning on landing new dash/RR releases and SDK updates soon.

PR #73, opened 2021-12-04
Add ignoreDisabled to @Config

ryleu:
Adds a field in `@Config` that lets you override the `@Disabled` annotation, adding the configuration variables to the dashboard regardless of whether `@Disabled` is present.

### DIFF
--- a/FtcDashboard/src/main/java/com/acmerobotics/dashboard/config/Config.java
+++ b/FtcDashboard/src/main/java/com/acmerobotics/dashboard/config/Config.java
@@ -11,7 +11,7 @@ import java.lang.annotation.Target;
  * <p>All public, static, non-final fields of the class will be automatically added as configuration
  * variables in the dashboard. When the user saves new values, these fields are correspondingly
  * updated. Classes annotated with {@link com.qualcomm.robotcore.eventloop.opmode.Disabled} are
- * ignored.
+ * ignored, unless you set {@link #ignoreDisabled} to `true`.
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.TYPE)
@@ -20,4 +20,9 @@ public @interface Config {
      * Name of this block of configuration variables. Defaults to the class's simple name.
      */
     String value() default "";
+
+    /**
+     * Ignore the @Disabled annotation. Effectively force-enables configuration.
+     */
+    boolean ignoreDisabled() default false;
 }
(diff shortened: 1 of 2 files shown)


ryleu:
Forgot to give my reasoning for adding this, so here it is:

We use a single variable to enable and disable vast swaths of op modes at once, aka the RR tuning op modes. The way we do this is by making use of the `@OpModeRegistrar` annotation and putting `@Disabled` on all of our tuning op mode classes. The issue with this is it disables the config too. By having this option be configurable, we could put that boolean into `@Config` and override `@Disabled`'s effect on the config.

This is one (admittedly odd) use-case, but I'm sure others will find more.

rbrott:
Thanks for the patch. I haven't used these APIs in a while, but I believe you can remove both `@Disabled` and `@Autonomous`/`@TeleOp` from each tuning op mode to solve your problem.

ryleu:
> Thanks for the patch. I haven't used these APIs in a while, but I believe you can remove both `@Disabled` and `@Autonomous`/`@TeleOp` from each tuning op mode to solve your problem. 

It throws an error if you try to register a class that does not have either `@Teleop` or `@Autonomous` from what I can remember. I can double check that, though.

ryleu:
Update on this: I tested it and it does work, but now there's no way to disable the config when I don't want the op modes to show up on the dash.

I guess a better plan of action would be to make a boolean you can pass in to turn it off. I'm going to close this PR and make a new one with that patch instead.

rbrott:
> I guess a better plan of action would be to make a boolean you can pass in to turn it off.

I'm assuming "it" here is a `@Config` annotation. How does this approach compare with commenting the `@Config` out?

rbrott:
Ah ok I'm finally processing the root problem. You can register configuration classes without reflection (a la `OpModeRegistrar`) with the more obscure tree-level API.
```java
FtcDashboard.getInstance().getConfigRoot()
  .putVariable("MyOpMode", ReflectionConfig.createVariableFromClass(MyOpMode.class));
```

ryleu:
> Ah ok I'm finally processing the root problem. You can register configuration classes without reflection (a la `OpModeRegistrar`) with the more obscure tree-level API.
> 
> ```java
> 
> FtcDashboard.getInstance().getConfigRoot()
> 
>   .putVariable("MyOpMode", ReflectionConfig.createVariableFromClass(MyOpMode.class));
> 
> ```
> 
> Update: After a batch of modifications, you should call `FtcDashboard#updateConfig()` to propagate the changes to the clients.

Awesome, thank you so much! Sorry for the confusion.

ryleu:
Okay, I got everything sorted out. I'm considering making a PR to the RR quickstart...

Thank you for your help!

PR #76, opened 2022-01-25
Fixed issue where dashboard crashes when double marked with config is infinite or NaN

Eeshwar-Krishnan:
someone had this issue thought I would pr it lol


rbrott:
Nice, this is indeed a common issue. This minimal patch gets the job done but doesn't really fit with my general philosophy of surfacing errors directly instead of silently proceeding. I'd prefer a change like one of the following (in order of increasing effort).
* Catch the error thrown by Gson and throw a new exception with a more instructive error message.
* Flag invalid values and display a per-variable message in lieu of edit controls.
* Extend controls to support special float values and transmit them faithfully (boxing if necessary).

PR #91, opened 2022-09-17
Fix gamepad axis mapping

z3db0y:
This issue has been bothering me for a long time and I have decided to open a PR to fix it:
The mapping on our team's controllers in the dashboard was off - left stick axes were mapped correctly but right_stick_x was mapped to the controller's right_stick_y and right_stick_y wasn't being mapped at all. Taking a look at the code, it was simply 2 typos.

### DIFF
--- a/FtcDashboard/dash/src/store/middleware/gamepadMiddleware.ts
+++ b/FtcDashboard/dash/src/store/middleware/gamepadMiddleware.ts
@@ -118,8 +118,8 @@ const extractGamepadState = (gamepad: Gamepad) => {
         // USB ID=24C6, PID=530A
         left_stick_x: cleanMotionValues(gamepad.axes[0]),
         left_stick_y: cleanMotionValues(gamepad.axes[1]),
-        right_stick_x: cleanMotionValues(gamepad.axes[3]),
-        right_stick_y: cleanMotionValues(gamepad.axes[4]),
+        right_stick_x: cleanMotionValues(gamepad.axes[2]),
+        right_stick_y: cleanMotionValues(gamepad.axes[3]),
 
         dpad_up: gamepad.buttons[12].pressed,
         dpad_down: gamepad.buttons[13].pressed,


rbrott:
Thanks for the patch. I'm not sure those two indices are actually typos, and it's impossible for me to test. I don't want to change the general case if possible, but I'm happy to add a new case for you. Can you give me the full name of the gamepad as reported by [this site](https://gamepad-tester.com/) as well as your browser and OS? For context, I'm hoping to fill out [this info](https://github.com/acmerobotics/ftc-dashboard/blob/c9a2b5618c9237ae8a3ce45ebf51f86b1b0326f4/FtcDashboard/dash/src/store/middleware/gamepadMiddleware.ts#L116-L118) for all gamepad changes going forward.

z3db0y:
# Sorry, for the late reply, but here is the test:
image
# Controller Used: official Xbox One Controller:
image

Windwoes:
One possible way to fix this would be to use WebHID to manually interpret the data reports for the FTC legal gamepad types.

NoahBres:
:(
CleanShot 2022-11-20 at 22 44 36@2x

rbrott:
To be clear: Chrome-only support for gamepads is reasonable. Doubly so if the implementation actually works without constant fiddling.

qwertychouskie:
Chrome-only would be a big bummer for our team.  Ideally there would at least be the old code path as a fallback, preferably with a config option in the dashboard to allow users to change their axis mappings.

NoahBres:
@qwertychouskie Do the other browsers work with gamepads now?

Glaferg:
Well... I don't know. A quick search at-Comp yielded [this Firefox library.](https://developer.mozilla.org/en-US/docs/Web/API/Gamepad_API/Using_the_Gamepad_API) for controller support, and Safari's apparently works, as detailed [here.](https://www.lambdatest.com/web-technologies/gamepad-support-on-safari-12) Incredibly eager to see one of the most helpful libs in FTC grow!!

qwertychouskie:
I will also mention I am able to use https://gamepad-tester.com/ just fine on Firefox+Linux.

NoahBres:
This isn't a problem with the gamepad API not working. The problem is that the gamepad API doesn't work in non-secure contexts. Local dash isn't HTTPS, thus not a secure context. More details on my investigation can be found here: https://github.com/acmerobotics/ftc-dashboard/issues/67

Those sites should work fine because they are served over HTTPS. The problem is that dash is not served over HTTPS.

z3db0y:
Lol you guys over-thought the issue, the API you're using is fine and well-supported, it's just that xbox controllers get mapped wrong. I asked another team about their dashboard sticks and they told me their right stick is also wrongly mapped. If a member of the repo has an xbox controller, I suggest testing it- here's proof of concept:

https://user-images.githubusercontent.com/69466123/222927690-754fa2cb-0769-46dd-8775-cacd53f1d82c.mp4

Cannot get axes[4] of a 4 element array xd

PR #94, opened 2022-09-26
Fixed linux PS4 dualshock touchpad support

MafteiAlbert-Alexandru:
Due to the way linux handles the touchpad of the PS4 DualShock controller(sees it as a pointing device), the gamepad doesn't work. This is a support fix, if the Team Code relies on the touchpad, there is no way to use it from the dashboard. With this fix, the touchpad button is ignored only on linux.

### DIFF
--- a/FtcDashboard/dash/src/store/middleware/gamepadMiddleware.ts
+++ b/FtcDashboard/dash/src/store/middleware/gamepadMiddleware.ts
@@ -144,36 +144,67 @@ const extractGamepadState = (gamepad: Gamepad) => {
         right_trigger: gamepad.buttons[7].value,
       };
     case GamepadType.SONY_DUALSHOCK_4:
-      return {
-        left_stick_x: cleanMotionValues(gamepad.axes[0]),
-        left_stick_y: cleanMotionValues(gamepad.axes[1]),
-        right_stick_x: cleanMotionValues(gamepad.axes[2]),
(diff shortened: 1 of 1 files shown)


FtcDashboard/dash/src/store/middleware/gamepadMiddleware.ts
```diff
@@ -144,36 +144,67 @@ const extractGamepadState = (gamepad: Gamepad) => {
...
-        dpad_up: gamepad.buttons[12].pressed,
-        dpad_down: gamepad.buttons[13].pressed,
-        dpad_left: gamepad.buttons[14].pressed,
-        dpad_right: gamepad.buttons[15].pressed,
-
-        a: gamepad.buttons[0].pressed,
-        b: gamepad.buttons[1].pressed,
-        x: gamepad.buttons[2].pressed,
-        y: gamepad.buttons[3].pressed,
+      if (gamepad.buttons[17] !== undefined)
```
rbrott, line 147: Can you store the object in a variable and then conditionally add `touchpad`? Hopefully that will emphasize the commonality between the two cases.

rbrott (commented):
Thanks for the patch! I just left a quick suggestion.

FtcDashboard/dash/src/store/middleware/gamepadMiddleware.ts
MafteiAlbert-Alexandru: Done

PR #97, opened 2022-10-11
Make implicitly used subdependencies explicit

nash-pillai:
This project uses the dependencies `lodash`, `prop-types`, and `react-resizable`, but does not include them in the package.json
There aren't any errors because they are subdependencies of explicitly declared dependencies, but this is bad practice and could lead to bugs in the future that are difficult to troubleshoot.


rbrott:
Thanks for the patch!

PR #114, opened 2022-11-29
Co-locate dependent components

NoahBres:
Prior:
- Component and the dependencies they used dispersed everywhere. Was fine initially but with the current number of components, you need to jump around a lot
CleanShot 2022-11-29 at 13 45 54@2x

After:
- Co-locate related components 
CleanShot 2022-11-29 at 14 38 59@2x


rbrott (commented):
Looks good, this is a great improvement. 

1. I'd prefer absolute imports everywhere. I figure this is just a lint setting? Is there a problem with this approach?

1. I see more tailwind lints in this diff. Is there a way to lint all the files and keep subsequent diffs a little cleaner?

NoahBres:
> I see more tailwind lints in this diff. Is there a way to lint all the files and keep subsequent diffs a little cleaner?

I did not even think to solve this problem. They support a prettier plugin now!

> I'd prefer absolute imports everywhere. I figure this is just a lint setting? Is there a problem with this approach?

Just to clarify, you'd like to enforce using `import foo from '@/bar'` everywhere?

rbrott:
Thanks for making the changes! `@/` is perfect.

PR #138, opened 2023-06-24
Field Versatility Updates

IronReignRobotics:
Features meant to enhance the versatility of field overlays to allow teams to visualize their custom coordinate systems and localization units. Can override the field background image, gridlines, rotation, origin, scaling, add text and control transparency better.


FtcDashboard/dash/src/components/views/FieldView/Field.js
```diff
@@ -144,18 +172,66 @@ export default class Field {
...
+            //first reset the scale to the default
+            this.adjustOrigin(this.ctx, defaultTransform, altOriginX, altOriginY, altRotation);
+            this.ctx.scale(op.scaleX, op.scaleY);
+            break;
+        case 'rotation':
+            altRotation = op.rotation;
+            this.adjustOrigin(this.ctx, defaultTransform, altOriginX, altOriginY, altRotation);
+            fieldAltImage
+            break;
+        case 'origin':
```
rbrott, line 227: Setting aside the current semantics, I wonder how you think mixing "set origin" with other transformation ops should work. I think it's easy for users to be confused here. Replacing this with a translate operation would bring the dash canvas API in line with the HTML canvas API here and other similar APIs that users may encounter elsewhere. (And maybe expose generic save/restore functionality?) For this purpose, I generally like the dash canvas API to be a thin veneer over the HTML canvas API.

FtcDashboard/dash/src/components/views/FieldView/Field.js
```diff
@@ -144,18 +172,66 @@ export default class Field {
     this.ctx.restore();
   }
 
+adjustOrigin(ctx, defaultTransform, altOriginX, altOriginY, altRotation){
+    ctx.setTransform(defaultTransform);
+    ctx.translate(altOriginX, altOriginY);
+    ctx.rotate(altRotation);
+}
+
+drawAltFieldImage(ctx, src, x, y, width, height, altX, altY, altWidth, altHeight, altOpaque){
```
rbrott, line 181: I'm a little suspicious of the image handling logic here. I would prefer a system that maintains a global list of loaded images and (to keep it simple) re-renders all of the fields when a new image loads (identity determined by filename). Then the drawing logic can just assume the image is loaded.

FtcDashboard/dash/src/components/views/FieldView/Field.js
```diff
@@ -110,9 +132,15 @@ export default class Field {
     this.ctx.save();
     this.ctx.globalAlpha = this.options.alpha;
     this.ctx.drawImage(fieldImage, x, y, width, height);
```
rbrott, line 134: Should users be able to disable this? I'm close to removing all of the special drawing here (alpha, image, grid lines) and instead making that a default prelude in the quickstart, in docs, etc.

FtcDashboard/dash/src/components/views/FieldView/Field.js
```diff
@@ -110,9 +132,15 @@ export default class Field {
...
+    if (fieldAltLoaded){
+        //this.ctx.drawImage(fieldAltImage, x, y, width, height);
+        if (this.altOpaque) this.ctx.globalAlpha = 1.0;
+        this.ctx.drawImage(fieldAltImage, x + (this.altIX/144 * width), y + (this.altIY/144 * height), this.altIWidth/144 * width, this.altIHeight/144 * width);
```
rbrott, line 138: Should users maybe be able to specify the size to render the image? We're so close to a generic image drawing capability which would be nice.

FtcDashboard/dash/src/components/views/FieldView/Field.js
```diff
@@ -144,18 +172,66 @@ export default class Field {
...
     this.ctx.lineCap = 'butt';
 
     this.overlay.ops.forEach((op) => {
       switch (op.type) {
+        case 'image':
+            this.drawAltFieldImage(this.ctx, op.src, x, y, width, height, op.x, op.y, op.width, op.height, op.opaque);
+            break;
+        case 'grid':
+            this.gridlinesHorizontal = (op.numHorizontal>2) ? op.numHorizontal : 2;
+            this.gridlinesVertical = (op.numVertical>2) ? op.numVertical : 2;
```
rbrott, line 215: If I'm reading this right, the settings here take an iteration to take effect.

I wonder if this operation should actually draw the grid and then users could choose to disable grids entirely, make multiple grids, etc. (This relates to the special drawing comment above.)

FtcDashboard/dash/src/components/views/FieldView/Field.js
```diff
@@ -165,6 +241,19 @@ export default class Field {
...
           this.ctx.lineWidth = op.width;
           break;
+        case 'text':
+            this.ctx.save();
+            this.ctx.rotate(op.theta);
+            this.ctx.font = op.font;
+            //have to flip y axis again temporarily or text will be mirrored
+            this.ctx.scale(1, -1);
+            if (op.stroke) {
+                this.ctx.strokeText(op.text, op.x, op.y)
```
rbrott, line 251: It seems like `(op.x, op.y)` are coordinates in the rotated frame which seems counterintuitive. (Also seems like the reflection should be applied immediately). I think translating, then rotating, then drawing text at the origin is more intuitive.

I'm not sure exactly how to handle the reflection. I think properly handling this requires saving the current transform, setting the "default transform", reflecting, drawing, and then restoring the saved transform.

rbrott (commented):
Thanks for the patch! I left a few comments aimed to iron out the canvas API, and then we can move onto the rest of the changes.

I'll have to think some on what to do with images. I like the convenience of prebundled images, but I don't want to inflate app size (and deploy times) too much.

FtcDashboard/dash/src/components/views/FieldView/Field.js
IronReignRobotics: >I generally like the dash canvas API to be a thin veneer over the HTML canvas API
Yup, we kinda detected that in your implementation and concur. That said, the separate translate and rotate (and scale) methods do exist in the canvas API and you were indeed using them if not exposing them. We just preserved the 1 to 1 mapping. While the transform is also available - and we had to use it for specific resetting operations - I think more ftcdash users will be comfortable with using the discrete translate and rotate (and scale) operations. Sure there are advanced high school coders that know how to prepare transform matrices, but I don't think that should be an expectation or requirement. If you would like us to add a user-available pass through CanvasOp for transforms, just say the word and we'll add it. That would allow skew operations. Just didn't see the need for it.

FtcDashboard/dash/src/components/views/FieldView/Field.js
IronReignRobotics: We were trying to keep it as simple as possible. We weren't going for a full sprite animation system. We just wanted to be able to fully override the default field image while mostly avoiding refactoring since we may not have complete understanding of all use cases. We wanted to keep it simply additive as much as possible. There may be other complications resulting from the original decision to trigger the rendering cascade based from the fieldImage.onload callback.

FtcDashboard/dash/src/components/views/FieldView/Field.js
IronReignRobotics: Yeah there were some default/setup operations that we had to work around or provide ways to override to get the functionality we needed.  If you as original author want to refactor this, you've all the power. It may not be necessary. This is still a more a tool for visual feedback and a very specific environment than it is a general purpose API. Sometimes the perfect is the enemy of the good enough?

FtcDashboard/dash/src/components/views/FieldView/Field.js
IronReignRobotics: I'm not sure what this is asking? We did add the ability to set the destination location and size in standard "inches". We definitely have use cases for that. See our updates to features.md and let us know if this doesn't make sense. Maybe I just misread your comment here.

FtcDashboard/dash/src/components/views/FieldView/Field.js
IronReignRobotics: it's true there is a one iteration lag due to this having to reset a default that gets applied before the CanvasOps are evaluated. As a matter of UX it's not really detectable. Again your call if you want to refactor the default/setup/prelude operations to make this cleaner.

They can disable the grid by setting the number of lines to 2 or below - in which case it just outlines the field. We considered multiple grids but couldn't come up with well described use cases for it.  One thing we came up with was a use case for multiple FieldViews. Not so much for actual fields, but we wanted another canvas view where we could draw the current articulation of our robot in an orthographic side view. Mostly to visualize planned vs. actual kinematics of our robot's arms. Alternately this can be squashed onto part of the default FieldView, but then we will be occluding other items.

FtcDashboard/dash/src/components/views/FieldView/Field.js
IronReignRobotics: This is one of those consistency things. Aside from the prelude operations, all mapped drawing ops get rendered in the most recently set transform (as arrived at by translate, rotate and scale). We decided to keep text rendering consistent with the other drawing ops. That certainly makes sense when drawing things like axes.

otoh, I can see your point as we humans have a strong affinity for page orientation when it comes to text and it's certainly possible to reset the transform around text ops if that's the way we need to go.

FtcDashboard/dash/src/components/views/FieldView/Field.js
rbrott: I see your concern about making non-additive changes, though I think combining alt field images with the current system for the default will give a simpler, more robust implementation at the end of the day. I'm happy to code up a patch proposal if that would help.

FtcDashboard/dash/src/components/views/FieldView/Field.js
rbrott: > I generally like the dash canvas API to be a thin veneer over the HTML canvas API
> > That said, the separate translate and rotate (and scale) methods do exist in the canvas API and you were indeed using them if not exposing them. We just preserved the 1 to 1 mapping.

I just want to clarify that I don't see the need to expose all canvas API methods through dash. I was mostly commenting that the "origin" op type does not translate obviously into a set of HTML canvas context calls. The implementation is "non-compositional" in that we have to store the coordinates and make adjustments to other transformation calls. 

I would propose that implementing a "translate" op type instead that just calls the canvas context `translate()` would more directly expose the HTML canvas API and preserve the transformation interactions in that API without making it more difficult for users to set an origin. My goal is to avoid making a custom transformation API different from HTML canvas / Cairo / Skia / etc. that users then have to learn.

There's a related argument here that the "scale" op type should apply a scale atop the existing transformation (i.e., simply call `scale()`).

> While the transform is also available - and we had to use it for specific resetting operations - I think more ftcdash users will be comfortable with using the discrete translate and rotate (and scale) operations. Sure there are advanced high school coders that know how to prepare transform matrices, but I don't think that should be an expectation or requirement.

I completely agree and didn't mean to imply that `transform()` should be exposed.

FtcDashboard/dash/src/components/views/FieldView/Field.js
rbrott: I think I'm the one that misread 😅 .

By "generic image drawing", I mean the ability to draw as many images in whatever operation order as the user wants.

FtcDashboard/dash/src/components/views/FieldView/Field.js
IronReignRobotics: I think I understand. We went off-book by giving the name "origin" to what is effectively just a translate command. We'll fix it.

FtcDashboard/dash/src/components/views/FieldView/Field.js
IronReignRobotics: Yes, it would help if you could tackle this bit. Our javascript/node fu is yellow belt at best.

FtcDashboard/dash/src/components/views/FieldView/Field.js
IronReignRobotics: Actually, looking again, it would probably make more sense for us to address this after your refactoring of the default scene setup operations since the way we implemented adjustOrigin() was very tied into the current approach and would have to change again afterward.

FtcDashboard/dash/src/components/views/FieldView/Field.js
IronReignRobotics: Yeah, this ends up looking like a sprite system, which could be cool. Does this mean we'd need to dynamically generate an onload callback for each new image added?

FtcDashboard/dash/src/components/views/FieldView/Field.js
IronReignRobotics: And now on triple thinking it, we've just starting refactoring to match translate instead of origin.

notjaisammpath: Just finished the refactoring.

FtcDashboard/dash/src/components/views/FieldView/Field.js
```diff
@@ -81,6 +95,14 @@ export default class Field {
...
       ops: [],
     };
+
+    this.altIX = 0;
+    this.altIY = 0;
+    this.altIWidth = 0;
+    this.altIHeight = 0;
+    this.altOpaque = false;
+    this.gridlinesVertical = 7;
+    this.gridlinesHorizontal = 7;
```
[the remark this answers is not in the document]
IronReignRobotics: yes, we are cleaning out the stuff that is no longer needed after your update

IronReignRobotics:
We think we've completed the requested changes and what has been discussed elsewhere. Ready for the PR to be reviewed again!

rbrott:
Thanks for the contribution! I'll do a little polishing on `master` and cut a new release soon.

IronReignRobotics:
Thank you for accepting our PR. We are eagerly looking forward to this
release. Our presentation is next Saturday and we'd prefer to test our
presentation against your normal gradle/maven distribution. Do you think it
will drop this week or do we need to plan on running the demo against our
fork?

PR #144, opened 2023-10-31
Add array config support

MafteiAlbert-Alexandru:
As it says, this config adds the ability to edit array values in the Config window of the dashboard. This is achieved by using different ValueProvider implementation(which accesses the values from the array) when creating nested CustomVariables, and using indices as the names for those variables.
This modification is safe, since resizing the array doesn't cause any exceptions, but should ideally somehow cause a refresh of the config variables.

### DIFF
--- a/TeamCode/src/main/java/org/firstinspires/ftc/teamcode/ArrayTestOpMode.java
+++ b/TeamCode/src/main/java/org/firstinspires/ftc/teamcode/ArrayTestOpMode.java
@@ -1,0 +1,34 @@
+package org.firstinspires.ftc.teamcode;
+
+import com.acmerobotics.dashboard.FtcDashboard;
+import com.acmerobotics.dashboard.config.Config;
+import com.acmerobotics.dashboard.telemetry.MultipleTelemetry;
+import com.qualcomm.robotcore.eventloop.opmode.LinearOpMode;
+import com.qualcomm.robotcore.eventloop.opmode.TeleOp;
+
+@SuppressWarnings("unused")
+@Config
+@TeleOp
+public class ArrayTestOpMode extends LinearOpMode {
+    public static class NestedClass {
+        public int a,b;
+    }
+    public static int[] array = new int[3];
+    public static NestedClass[] innerArray = new NestedClass[] { new NestedClass(), new NestedClass(), new NestedClass()};
+    @Override
+    public void runOpMode() {
+        telemetry = new MultipleTelemetry(telemetry, FtcDashboard.getInstance().getTelemetry());
+
+        waitForStart();
+        array=new int[2];
+        while(opModeIsActive())
+        {
+            telemetry.addData("array0", array[0]);
+            telemetry.addData("array1", array[1]);
+            telemetry.addData("innerArray0A", innerArray[0].a);
+            telemetry.addData("innerArray1B", innerArray[1].b);
+            telemetry.addData("innerArray2A", innerArray[2].a);
+            telemetry.update();
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


TeamCode/src/main/java/org/firstinspires/ftc/teamcode/ArrayTestOpMode.java
```diff
@@ -0,0 +1,34 @@
...
+        public int a,b;
+    }
+    public static int[] array = new int[3];
+    public static NestedClass[] innerArray = new NestedClass[] { new NestedClass(), new NestedClass(), new NestedClass()};
+    @Override
+    public void runOpMode() {
+        telemetry = new MultipleTelemetry(telemetry, FtcDashboard.getInstance().getTelemetry());
+
+        waitForStart();
+        array=new int[2];
```
rbrott, line 23: Seems like this shouldn't be allowed?

rbrott (commented):
Good stuff. I'll check it out locally and play around.

rbrott:
I put together a test config class.
```java
package com.acmerobotics.dashboard.testopmode;

public class TestConfig {
    public static int[] NUM_ARRAY = {1, 2, 3};
    public static int[] EMPTY_ARRAY = {};
    public static int[] NULL_ARRAY = null;

    public static int[][] NESTED_ARRAY = {{1, 2}, {3, 4}};
    public static int[][] NESTED_EMPTY_ARRAY = {{}, {}};

    public static int[][] RAGGED_ARRAY = {{1, 2}, {3, 4, 5}};

    public static class NestedClass {
        public int[] numArray = {1, 2, 3};
        public int[] emptyArray = {};
        public int[] nullArray = null;

        public int[][] nestedArray = {{1, 2}, {3, 4}};
        public int[][] nestedEmptyArray = {{}, {}};

        public int[][] raggedArray = {{1, 2}, {3, 4, 5}};
    }
    public static NestedClass NESTED_CLASS = new NestedClass();
}
```

And this PR seems to handle the one-dimensional arrays great (:tada:) and recurses infinitely on the two-dimensional ones.

rbrott:
I'm testing with `TestServer` and the following patch
```diff
diff --git a/DashboardCore/src/test/java/com/acmerobotics/dashboard/TestDashboardInstance.java b/DashboardCore/src/test/java/com/acmerobotics/dashboard/TestDashboardInstance.java
index 3fb819e..2a3fe56 100644
--- a/DashboardCore/src/test/java/com/acmerobotics/dashboard/TestDashboardInstance.java
+++ b/DashboardCore/src/test/java/com/acmerobotics/dashboard/TestDashboardInstance.java
@@ -5,11 +5,13 @@ import com.acmerobotics.dashboard.RobotStatus;
 import com.acmerobotics.dashboard.SendFun;
 import com.acmerobotics.dashboard.SocketHandler;
 import com.acmerobotics.dashboard.config.ValueProvider;
+import com.acmerobotics.dashboard.config.reflection.ReflectionConfig;
 import com.acmerobotics.dashboard.message.Message;
 import com.acmerobotics.dashboard.message.redux.InitOpMode;
 import com.acmerobotics.dashboard.message.redux.ReceiveOpModeList;
 import com.acmerobotics.dashboard.message.redux.ReceiveRobotStatus;
 import com.acmerobotics.dashboard.telemetry.TelemetryPacket;
+import com.acmerobotics.dashboard.testopmode.TestConfig;
 import com.acmerobotics.dashboard.testopmode.TestOpMode;
 import com.acmerobotics.dashboard.testopmode.TestOpModeManager;
 
@@ -181,6 +183,9 @@ public class TestDashboardInstance {
             }
         });
 
+        core.withConfigRoot(configRoot ->
+            configRoot.putVariable("TestConfig", ReflectionConfig.createVariableFromClass(TestConfig.class)));
+
 
         try {
             server.start();
```
Maybe that can help speed up your workflow.

MafteiAlbert-Alexandru:
Worked on nested/multidimensional arrays and they work now.

TeamCode/src/main/java/org/firstinspires/ftc/teamcode/ArrayTestOpMode.java
MafteiAlbert-Alexandru: This was an exception safety test.

rbrott (approved):
Works great, thanks for the contribution!

PR #146, opened 2023-12-05
Rewrite the camera section of documentation

KennanHunter:
Updated for Vision Portal and put a bit more of a description for traditional EasyOpenCV support.

Looks like the line endings got a bit messed up, perhaps one of you could format it with your preferences?

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -6,7 +6,7 @@ layout: default
 
 ## Telemetry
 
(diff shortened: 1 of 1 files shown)


rbrott (commented):
Nice, thanks for the patch. I forgot that the camera section still reference the Vuforia sample.

PR #149, opened 2023-12-11
Fix link to Node version

BD103:
This is a small fix that corrects the line number of the link. I also made the link target the master branch, in case this numbers changes in the future. That also means that the number will be incorrect again if a line is added or removed.


rbrott:
Great, thanks for the patch.

PR #150, opened 2023-12-11
Add support for more primitives

### LINKED ISSUE #148: Configuration does not support long numbers
When I try editing [`long`](https://docs.oracle.com/javase/8/docs/api/java/lang/Long.html) variables through FTC Dashboard, they are treated as empty dropdowns instead of actual numbers.

image

```kotlin
@Config
object AprilVision {
    @JvmField
    var OPTIMUM_EXPOSURE: Long = 1

    @JvmField
    var OPTIMUM_GAIN: Int = 255
}
```

I am using Kotlin, but I can also reproduce it in Java if requested. I theorize the reason this is happening is because the JVM treats `long` types as two `int`s, but FTC Dashboard's serialization can't handle it.

BD103:
This is a draft PR that adds support for more Java primitive types. Closes #148.

# Todo

- [x] `long`
- [x] `float`

# Optional

- [ ] `short`
- [ ] `byte`
- [ ] `char`

### DIFF
--- a/DashboardCore/src/main/java/com/acmerobotics/dashboard/config/reflection/ReflectionConfig.java
+++ b/DashboardCore/src/main/java/com/acmerobotics/dashboard/config/reflection/ReflectionConfig.java
@@ -33,6 +33,7 @@ public class ReflectionConfig {
         switch (type) {
             case BOOLEAN:
             case INT:
+            case LONG:
             case DOUBLE:
             case STRING:
             case ENUM:
@@ -88,6 +89,7 @@ public class ReflectionConfig {
         switch (type) {
             case BOOLEAN:
             case INT:
+            case LONG:
             case DOUBLE:
             case STRING:
             case ENUM:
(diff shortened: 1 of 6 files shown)


rbrott:
> @rbrott this is now ready as a minimum viable product.

Great work, thanks for the patch.

>     * Should I add support for `short`, `byte`, and / or `char`?  
>       * I feel they're less common, so I don't know if it's worth implementing it (yet).

I agree, those can wait.

>     * Do you want me to remove the [test configuration](https://github.com/acmerobotics/ftc-dashboard/pull/150/files#diff-f2ed284d340b830a45aacf109e4d4895ebbc84710b1868472bf3724f6618e382) or is it fine to keep?

Fine to leave it.

>     * Should the web client check that the number is less than 263? It seems to silently error out when a too large number is given.

I don't think we need to address proper bounds right now. It would be nice to add full support for longs in the future, but this is a useful improvement as-is.

BD103:
@rbrott would you mind publishing a new release with this pull request? My team needs the feature, but I'm unsure of how to depend on a Git repository.

PR #152, opened 2024-05-08
Add public getter for if the dashboard is active

GraysonTinker:
(First time making a PR, so please let me know if I made a mistake somewhere/should change something.)

This is just a simple change to let users check if the dashboard is enabled from code. One use case, and the situation I found myself wanting this for, is to allow selective disabling of elements which are designed for testing and are useless without the dashboard (specifically, a custom speed setting for tuning set within the web interface). However, I'm sure there are other potential use cases.


rbrott:
Thanks for the patch! Sorry I missed this and took so long to act on it.

PR #158, opened 2024-11-06
Ensure op mode name for logging does not get overwritten by $Stop$Robot

monkbroc:
When using the logging view to save a CSV, the file name always starts with `$Stop$Robot` no matter which op mode was run.

This PR ensures that the name of the current op mode for the logging view does not get overwritten by `$Stop$Robot`.

Last downloaded file in this screenshot is after the change, previous ones are before.
image


rbrott:
Nice, thanks for the patch!

Issue #50, opened 2021-03-02
Changes Popup


Jarvis7529:
I was wondering if you can add a popup with here are the configs that changed since you started so you could make sure you update the real values in the code?

NoahBres:
I actually was thinking about this a while ago but don't really have an idea for how the UI would be implemented. Firefox's dev tools actually shows the modified css values, which was the impetus for the idea. However, I am not entirely sure how to translate that to Dashboard's UI.

Jarvis7529:
My thought is have a button up by the save disk that would popup a model window showing original and changed values. Maybe show all values with some JS that would allow you to show all or just changes,

PR #161, opened 2025-02-05
Dashboard Configuration Variable Highlighting

aaditn:
Marking dashboard config variables if they've been changed from their uploaded value.


rbrott:
Thanks for the patch!

> Marking dashboard config variables if they've been changed from their uploaded value.

I take the uploaded value to mean the value that was in the file deployed to the bot. The client needs more information from the backend to provide that value and shouldn't assume that the first config received by the client contains the deployed values.

Also you should use `yarn format` to format `client` changes. It's also great for graphical changes if you can post a screenshot of the new UI.

mxtmx:
This would also resolve #50.

mxtmx:
Merge conflicts have been resolved

mxtmx:
Sorry for the delay, just fixed!

mxtmx:
All those changes should be made! I also ran `yarn format`.
Screenshot 2026-03-22 at 11 56 05 AM
Screenshot 2026-03-22 at 11 56 00 AM

PR #163, opened 2025-02-05
Update FTC and Android SDK

mxtmx:
Update to FTC SDK 10.1.1 and Android SDK 35.


qwertychouskie:
Should probably update to 10.2 as it contains a fix for RUN_TO_POSITION potentially hard-crashing the hubs and requiring a power cycle.

mxtmx:
Yep that makes a lot of sense. Thank you!

rbrott:
Thanks for the patch! I actually prefer to update myself just to make sure the repository tracks upstream as closely as possible. I try to stay away from upgrading anything past the versions upstream.

PR #168, opened 2025-03-30
Add Hardware View

mxtmx:
Hardware View is a manually activated view that displays information about motors and servos. It also allows you to change certain values about the motors and servos for testing in real life.

Hardware View allows you to view:
- Current draw on individual motors
- Encoder position for each motor
- Motor power
- Servo position
- The port and device (control hub or expansion hub) that the motors and servos are on

It also lets you make the following changes to actuate motors and servos in real life:
- Change motor power
- Change servo position
- Utilize all the run modes for motors (run to position, run using encoder, etc.)

The view is manually activated to avoid unnecessary reads.

Demo Videos (we fixed the current rounding):
https://drive.google.com/file/d/1dRQ2LkXtBon-k48-pxTyeQvAAIsr0ulV/view?usp=share_link
https://drive.google.com/file/d/1KhRICWt38M8PwvKTI8D11mElkGRjY6WR/view?usp=share_link

PR #171, opened 2025-04-14
Add ability to switch the current hardware configuration from the web

### LINKED ISSUE #106: [Feature] Hardware Map editing

GraysonTinker:
This PR adds the ability for user to change what predefined hardware configuration is active through the Dashboard. Configurations must still be defined through the Driver Station.

Because it does not allow arbitrary edits to be made to the configuration, it does not resolve #106, although I would love for that to be implemented and I hope to work on that in the near(ish) future as an expansion of this system.

I plan to test this on actual hardware tomorrow; until then it is marked as a draft as there are certain concerns I want to test through first (primarily the config list getting out date). In the meantime, I would love feedback as I'm sure there are some problems/best-practices I missed :D.

image

### DIFF
--- a/FtcDashboard/src/main/java/com/acmerobotics/dashboard/FtcDashboard.java
+++ b/FtcDashboard/src/main/java/com/acmerobotics/dashboard/FtcDashboard.java
@@ -667,12 +667,14 @@ public class FtcDashboard implements OpModeManagerImpl.Notifications {
 
                     activeOpMode.with(o -> {
                         // Don't allow changing the config unless stopped. Who knows what undefined behavior that would cause
-                       if(o.status != RobotStatus.OpModeStatus.STOPPED) return;
+                       if(o.status != RobotStatus.OpModeStatus.STOPPED &&
+                               !opModeManager.getActiveOpModeName().equals("$Stop$Robot$")) return;
 
                        hardwareConfigList.with(l -> {
                            hardwareConfigManager.setActiveConfig(false, l.get(hardwareConfigName));
                        });
                     });
+                    break;
                 }
                 default: {
                     Log.w(TAG, "Received unknown message of type " + msg.getType());
--- a/FtcDashboard/src/main/java/com/acmerobotics/dashboard/FtcDashboard.java
+++ b/FtcDashboard/src/main/java/com/acmerobotics/dashboard/FtcDashboard.java
@@ -46,6 +46,8 @@ import fi.iki.elonen.NanoHTTPD;
 import fi.iki.elonen.NanoWSD;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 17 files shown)


GraysonTinker:
One other note - from the quick glance I took at the additions in #168, this PR doesn't seem conflict and works by a different method. However, if that PR gets merged and changes on this end do turn out to be necessary please let me know and I'd be happy to make them.

GraysonTinker:
Still a few kinks to work out here, I believe primarily on the client-side. My biggest concern turned out to be a non-issue.

GraysonTinker:
Okay, I've worked out all the problems I can find so I'm unmarking this as a draft. 

I will say that in order to restart the robot after a configuration change (necessary for the change to have any effect) we have to do a bit of reflection. Unfortunately I think this is the best method available to us without a new SDK API call, though of course I'd love to be proven wrong.

It's obviously not the best thing in the world but I think it makes sense here given the value of having this feature and others down-the-line which would also be dependent on it.

FtcDashboard/src/main/java/com/acmerobotics/dashboard/FtcDashboard.java
```diff
@@ -208,6 +216,9 @@ protected NanoWSD.WebSocket openWebSocket(NanoHTTPD.IHTTPSession handshake) {
     private TextView connectionStatusTextView;
     private LinearLayout parentLayout;
 
+    private RobotConfigFileManager hardwareConfigManager = new RobotConfigFileManager();
+    private final Mutex<Map<String, RobotConfigFile>> hardwareConfigList = new Mutex<>(new TreeMap<>());
```
rbrott, line 220: I assume you're using `TreeMap` for sorted keys. If that's the case, I'd rather the type on the left be a `SortedMap`.

FtcDashboard/src/main/java/com/acmerobotics/dashboard/FtcDashboard.java
```diff
@@ -628,6 +664,43 @@ protected void onMessage(NanoWSD.WebSocketFrame message) {
...
                     break;
                 }
+                case SET_HARDWARE_CONFIG: {
+                    String hardwareConfigName = ((SetHardwareConfig) msg).getHardwareConfigName();
+
+                    activeOpMode.with(o -> {
+                        // Don't allow changing the config unless stopped. Who knows what undefined behavior that would cause
+                       if(o.status != RobotStatus.OpModeStatus.STOPPED &&
+                               !opModeManager.getActiveOpModeName().equals("$Stop$Robot$")) return;
+
```
rbrott, line 674: two nits
* I prefer to always have curly braces even for a single line
* It would be great to use `OpModeManager.DEFAULT_OP_MODE_NAME`

rbrott: Shouldn't this check apply to the robot restart as well?

client/src/components/ConfigurableLayout/ConfigurableLayout.tsx
```diff
@@ -41,6 +41,7 @@ import CreateIconURL from '@/assets/icons/create.svg';
 
 import { colors } from '@/hooks/useTheme';
 import { useTheme } from '@/hooks/useTheme';
+import HardwareConfigView from '../views/HardwareConfigView';
```
rbrott, line 44: nit: Can you use an absolute path and put this with the views further up in this file?

rbrott (commented):
This patch is in great shape. Sorry I took so long to review.

rbrott:
Great, thanks. Would you mind merging in master and resolving the conflicts?

rbrott:
I was checking this out locally and it's working great. My one piece of feedback is that it would be nice to have some differentiation between "user just picked a new config and it's being applied on the RC" and "the selected config is active" (maybe something like the config variable asterisks? doesn't have to be big and obtrusive).

PR #174, opened 2025-04-18
Added support for Logitech F310 controller

ngerow2025:
I noticed that the reason that my controller did not work for me was that the Logitech F310 controller was simply not supported in the code so I added support and tested, it worked on my machine and it ended up very similar to the other controllers so I believe that everything should be good?

### DIFF
--- a/client/src/store/middleware/gamepadMiddleware.ts
+++ b/client/src/store/middleware/gamepadMiddleware.ts
@@ -179,6 +179,38 @@ const extractGamepadState = (gamepad: Gamepad) => {
         return state;
       }
       break;
+    case GamepadType.LOGITECH_F310: {
+      let res =  {
+        left_stick_x: cleanMotionValues(gamepad.axes[0]),
+        left_stick_y: cleanMotionValues(gamepad.axes[1]),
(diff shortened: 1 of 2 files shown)


client/src/store/middleware/gamepadMiddleware.ts
```diff
@@ -179,6 +179,38 @@ const extractGamepadState = (gamepad: Gamepad) => {
         return state;
       }
       break;
+    case GamepadType.LOGITECH_F310: {
+      let res =  {
```
rbrott, line 183: I'd slightly prefer to return this object literal directly instead of creating an intermediate variable.

And you don't need `break;` after a `return ...;`.

rbrott (commented):
Thanks for the patch!

rbrott:
It would also be great if you could include a comment describing as much about your gamepad as you can. Here's an example from the code:
https://github.com/acmerobotics/ftc-dashboard/blob/a2ba8c5c77a59d8ca8333f618f0db7d8a7a3e9a6/client/src/store/middleware/gamepadMiddleware.ts#L116-L118

Gamepad mapping changes like this are hard for me to evaluate and it helps to have as much information as possible if issues come up in the future.

PR #177, opened 2025-05-28
Added support for Sony DualSense controller

Braedeng0:
I noticed my PS5 DualSense controller didn't work with FTCDashboard due to it not being listed as a supported controller in the code. By adding it and treating it the same as a PS4 DualShock controller, it seems to work fine with the correct button mapping.


rbrott:
Thanks!

PR #182, opened 2025-06-19
Add ability to stream Limelight 3As to the dashboard

### LINKED ISSUE #180: Limelight 3A Camera Stream
Don't know if it's possible, but since the config interface shows its stream certainly there's a way to send it back to the dashboard, right?
Would be appreciated.
-2845

GraysonTinker:
This PR adds the ability for Limelight 3As to be livestreamed to the web dashboard through a new overload of the startCameraStream() function.

While active, an HTTP connection is made, the currently-active pipeline's camera stream is parsed, and data is sent to clients. In the event of an error with the frame data, the stream will attempt to restart itself. If the active pipeline is switched by the OpMode, the stream will update accordingly.

Screenshot_20250619-095807

Note: This PR includes a minimum version bump to 10.0.0, the first version the Limelight3A class was added. This was done to improve the user experience; the IP address of the Limelight is needed to connect to the stream, and it's relatively easy to pull it from the object. However there are two alternatives if this is undesirable:
- Have the user manually pass the IP address. This can work, but it's challenging to acquire without a Driver Station and if it were to change (it's unclear how likely this is) it would require a rebuild.
- Parse the active configuration to determine the IP address. While this is viable, it raises some questions for future improvements like multi-camera streams or support for new products that work in a similar fashion.

Resolves #180.


FtcDashboard/src/main/java/com/acmerobotics/dashboard/FtcDashboard.java
```diff
@@ -526,6 +534,133 @@ public void accept(Bitmap value) {
...
 
+    private class LimelightCameraStreamRunnable implements Runnable {
+        private HttpURLConnection limelightConnection;
+        private BufferedInputStream byteStream;
+        private double maxFps;
+        private String ipAddress;
+
+        private LimelightCameraStreamRunnable(String ipAddress, double maxFps) {
+            try {
+                this.limelightConnection = (HttpURLConnection) new URL("http://" + ipAddress + ":5802").openConnection();
```
rbrott, line 545: Is there documentation or code that references this number? It would be nice to have a reference on the protocol.

FtcDashboard/src/main/java/com/acmerobotics/dashboard/FtcDashboard.java
```diff
@@ -526,6 +534,133 @@ public void accept(Bitmap value) {
...
+        }
+
+        private void restartStream() throws InterruptedException {
+            limelightConnection.disconnect();
+            Thread.sleep(500);
+            stopCameraStream();
+            Thread.interrupted(); // clear interrupt state
+
+            cameraStreamExecutor = ThreadPool.newSingleThreadExecutor("camera stream");
+            cameraStreamExecutor.submit(new LimelightCameraStreamRunnable(ipAddress, maxFps));
```
rbrott, line 660: I don't like the idea of creating a new thread -- can the same thread clean up its resources and open a new connection?

FtcDashboard/src/main/java/com/acmerobotics/dashboard/FtcDashboard.java
```diff
@@ -526,6 +534,133 @@ public void accept(Bitmap value) {
...
+                return;
+            }
+            while (!Thread.currentThread().isInterrupted()) {
+                try {
+                    final long timestamp = System.currentTimeMillis();
+
+                    if (core.clientCount() == 0) {
+                        Thread.sleep(250);
+                        continue;
+                    }
```
rbrott, line 572: Should the stream be closed in this case? Aren't frames still being sent by the server even if the client isn't reading them? I have a similar concern about the mechanism used to limit FPS.

FtcDashboard/src/main/java/com/acmerobotics/dashboard/FtcDashboard.java
```diff
@@ -526,6 +534,133 @@ public void accept(Bitmap value) {
...
+
+                    /*
+                     * Implementation note: We use our own parsing function as opposed to a more standard
+                     * Reader because we need access to both the text data (headers) and binary data
+                     * (image) at different points. Most (all?) Readers internally use buffers which
+                     * take more bytes from the stream than we read, causing binary data to be swallowed
+                     * into an endless void we can't retrieve from.
+                     */
+
+                    byteStream.skip(64); // Go to the start of the Content-Length integer. 60 chars + 4 from 2x \r\n
```
rbrott, line 594: Can you implement something more robust here? I'm nervous about relying on this exact length being correct.

FtcDashboard/src/main/java/com/acmerobotics/dashboard/FtcDashboard.java
GraysonTinker: There's no modern documentation for 5802 beyond minor references in changelogs and it's use in the web interface. I did find some legacy documentation that references it [here](https://github.com/LimelightVision/LimelightDocs/blob/master/docs/grip_software.rst): 

> Limelight has an additional video stream on port 5802 which can be accessed primarily for use with GRIP or other applications like it. This video stream is uncompressed (or very lightly compressed) and has no cross-hair or other overlays drawn on the image.

There's also the slightly-better-documented port 5800 that gets a reference in the [3A Quick-Start page](https://docs.limelightvision.io/docs/docs-limelight/getting-started/limelight-3a#for-frc-1):

> // (robotIP):5800 will now point to a Limelight3A's (id 0) video stream:

I don't have an image on hand but this version has a crosshair, FPS counters in the top left, and based on the above is presumably more compressed. Anecdotally I believe it also has a lower resolution (though I could be mistaken). If bandwidth is a concern then it would be easy to switch over to that.

As for protocol, it's a standard [MJPEG](https://en.wikipedia.org/wiki/Motion_JPEG) stream (or at least standard enough to be parseable by browsers). I don't believe they've done anything special so any standard documentation available online should be accurate. (That being said I did struggle to find one to double-check.)

I could include the legacy documentation link in the code if you'd like?

rbrott: It would be great if you could put the quote you pulled from the legacy docs and the link in a comment.

FtcDashboard/src/main/java/com/acmerobotics/dashboard/FtcDashboard.java
```diff
@@ -526,6 +535,151 @@ public void accept(Bitmap value) {
...
+                try {
+                    if (core.clientCount() == 0) {
+                        if (limelightConnection != null) { // Close connection to avoid backlog of frames
+                            limelightConnection.disconnect();
+                            limelightConnection = null;
+                            byteStream = null;
+                        }
+                        Thread.sleep(250);
+                        continue;
+                    } else if (limelightConnection == null) {
```
rbrott, line 593: Should disconnects from this bit of code count towards `reinitCount`?

FtcDashboard/src/main/java/com/acmerobotics/dashboard/FtcDashboard.java
```diff
@@ -526,6 +535,151 @@ public void accept(Bitmap value) {
...
+                        initialize(); // Wake back up
+                    }
```
rbrott, line 595: Isn't it still possible for `initialize()` to fail and leave `limelightConnection` null here?

FtcDashboard/src/main/java/com/acmerobotics/dashboard/FtcDashboard.java
```diff
@@ -526,6 +535,151 @@ public void accept(Bitmap value) {
...
+                        RobotLog.ee(TAG, "Invalid/Unexpected Limelight JPEG data (failed at start); restarting stream");
+                        byteStream.reset();
+                        // Can't just continue because it will parse binary data as headers next loop
+                        // Instead, we'll live with the dropped frames and just restart the stream
+                        initialize();
+                        if (limelightConnection != null) {
+                            continue;
+                        } else {
+                            return;
+                        }
```
rbrott, line 636: Should this just null the connection, continue, and let the logic earlier in the loop establish a new connection?

FtcDashboard/src/main/java/com/acmerobotics/dashboard/FtcDashboard.java
```diff
@@ -526,6 +535,151 @@ public void accept(Bitmap value) {
...
+                    String contentLength = readLine(byteStream); // Get the Content-Length header.
+                    String num = contentLength.replaceAll("[^0-9]", ""); // Filter to the integer
+                    int length = Integer.parseInt(num);
+                    readLine(byteStream);
+                    readLine(byteStream); // Go to start of binary
+
+                    byteStream.mark(2);
+                    if(byteStream.read() != 0xFF || byteStream.read() != 0xD8) { // All JPEGs start with FFD8; quick sanity-check
+                        RobotLog.ee(TAG, "Invalid/Unexpected Limelight JPEG data (failed at start); restarting stream");
+                        byteStream.reset();
```
rbrott, line 628: nit: superfluous reset since the connection and stream will be discarded

FtcDashboard/src/main/java/com/acmerobotics/dashboard/FtcDashboard.java
```diff
@@ -526,6 +535,151 @@ public void accept(Bitmap value) {
...
+
+                    // All JPEGs end with 0xFF and 0xD9; sanity check.
+                    if (out[length - 2] != (byte) 0xFF || out[length - 1] != (byte) 0xD9) {
+                        RobotLog.ee(TAG, "Invalid/Unexpected Limelight JPEG data (failed at end); restarting stream.");
+                        initialize(); // interrupts this thread and makes a new one
+                        if (limelightConnection != null) {
+                            continue;
+                        } else {
+                            return;
+                        }
```
rbrott, line 656: ditto on null the connection and continue

FtcDashboard/src/main/java/com/acmerobotics/dashboard/FtcDashboard.java
GraysonTinker: I don't think so, as that is a state which can be entered as normal behavior. (Avoiding it being incremented is why that code cleans up itself instead of leaving it to `initialize()`.) The name of the variable is a bit confusing though, so I'll rename it to make it more clear that it is tracking failures in the connection rather than all re-initializations.

FtcDashboard/src/main/java/com/acmerobotics/dashboard/FtcDashboard.java
GraysonTinker: I haven't implemented this exactly as you've suggested, but I have changed things so it reuses the existing logic. Part of the issue with simply nulling `limelightConnection` is that `initialize()` uses it to determine if an error state has been reached, meaning disconnection needs to be handled and reinitCount needs to be incremented. Instead I've resolved it with a new helper function for cleaning up the connection and extra state-tracking.

FtcDashboard/src/main/java/com/acmerobotics/dashboard/FtcDashboard.java
```diff
@@ -526,6 +535,163 @@ public void accept(Bitmap value) {
...
+                return true;
+            } catch (Exception e) {
+                RobotLog.ee(TAG, "Failed to connect to the Limelight camera stream.");
+                limelightConnection.disconnect();
+                limelightConnection = null;
+                return false;
+            }
+        }
+
+        private boolean needsInit = true;
```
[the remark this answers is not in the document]
GraysonTinker: I replied to this on the original comment but I think it got hidden by the resolution. Here's it copy-pasted:

I haven't implemented this exactly as you've suggested, but I have changed things so it reuses the existing logic. Part of the issue with simply nulling limelightConnection is that initialize() uses it to determine if an error state has been reached, meaning disconnection needs to be handled and reinitCount needs to be incremented. Instead I've resolved it with a new helper function for cleaning up the connection and extra state-tracking.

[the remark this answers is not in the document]
GraysonTinker: Sorry, I misunderstood your first comment. I believe you are correct and I'll remove it.

rbrott:
Thanks for bearing with the review! I'm pretty happy with the current patch.

PR #186, opened 2025-08-10
Support the standard gamepad mapping

### LINKED ISSUE #48: Defaulting to standard gamepad mapping when controller is unknown
I think this would require some testing with non-ftc gamepads but it would be neat to just default to the [standard gamepad layout](https://w3c.github.io/gamepad/#remapping) when the ID isn't known, enabling the use of non-FTC legal gamepads for use at home or something (there have been a non-zero amount of such requests, primarily Dualshock 5 and Xbox One controllers but not really important enough to spend significant time on imo).
With PR #47, the Dualshock 4 and Xbox 360 seem to have identical button layouts. I'm assuming that most modern gamepads will share the standard w3c spec.

GraysonTinker:
This PR improves compatibility for gamepads across browsers by making use of the 'standard' [W3C Gamepad mapping](https://www.w3.org/TR/gamepad/#remapping). All browsers which implement the spec are 'recommended' to convert non-standard orderings to the standard mapping, which means making use of this mapping is the only way to guarantee compatibility long-term as browsers perform these conversions for existing controllers over time. Any controller-specific change we make (excluding the handling of additional buttons) would eventually become obsolete as the controller is handled by the browser itself. 

The existing `XBOX_360` case varies only in a way I believe to be a bug (see below) from the Standard mapping, so I have opted to rename it to `STANDARD` and use it for this case. The existing checks for the category have been kept, so there is no loss in detected compatibility. These controllers are typically reported as simply 'xinput' on Firefox rather than the full name given in Chromium, which means that the same controller could be considered supported on one browser while being seen as not supported on another despite working on both. I've added an additional check to rectify this, but I believe that both of these checks can be removed entirely, or at least simplified to only the 'xinput' search.

This PR also fixes the issue outlined in #91. The reference gamepad shown in that PR, the gamepad used for originally testing the `XBOX_360` case (as outlined in the code), and both of the gamepads I am using for testing (and have successfully reproduced the issue on) all report a standard mapping, which I believe shows that non-conformance to the W3C spec is a bug. Additionally, the comment describes the case as the same as the `SONY_DUALSHOCK_4` case, but the `XBOX_360` case varies in this way. And #47, the original PR adding proper Xbox controller support, makes clear that the intended mapping was entirely standard.

One point I wasn't sure about is the reason that `guide` is not correctly mapped to the buttons array. Given this button is both a part of the spec and used within `SONY_DUALSHOCK_4`, I believe it should be possible to begin handling the button normally. (See also #48, which briefly mentions this issue.) If this change can be made, the `SONY_DUALSHOCK_4` and `STANDARD` cases will be practically identical except for the non-standard touchpad button, and I believe it would be possible (and debateably, desireable) to merge the cases into a single block. Let me know your thoughts on this.

Given the mapping used in #174 is identical to the one used here, I believe this PR will effectively provide support for the controller and *should* be capable of closing it.

I've tested this on Windows 11 with Chrome v138, Edge v139, and Firefox v135 with both an Xbox 360 controller and a generic XInput controller with Xbox buttons. I'm marking this as a draft for now to allow for feedback; next week I will test on Linux with Chromium/Firefox with both of those controllers and an additional Playstation 4 controller. I don't expect any issues.

One other side-note: It may be a good idea to re-test the Logitech Dual Action controller to see if browsers now support automatically remapping it to a standard mapping. If so, as noted above the current implementation is likely no longer functional.

Closes #48.

The standard button mapping

### DIFF
--- a/FtcDashboard/src/main/java/com/acmerobotics/dashboard/FtcDashboard.java
+++ b/FtcDashboard/src/main/java/com/acmerobotics/dashboard/FtcDashboard.java
@@ -1408,35 +1408,40 @@ public class FtcDashboard implements OpModeManagerImpl.Notifications {
     }
 
     public static void copyIntoSdkGamepad(ReceiveGamepadState.Gamepad src, Gamepad dst) {
-        dst.left_stick_x = src.left_stick_x;
-        dst.left_stick_y = src.left_stick_y;
-        dst.right_stick_x = src.right_stick_x;
-        dst.right_stick_y = src.right_stick_y;
+        // We need to copy from an intermediate so the SDK can handle the rising/falling edge detection
+        // Also, doing it like this means the SDK handles equivalencies between
+        // standard and Playstation buttons (i.e. converting A -> Cross and vice versa)
+        Gamepad intermediate = new Gamepad();
+        intermediate.left_stick_x = src.left_stick_x;
+        intermediate.left_stick_y = src.left_stick_y;
+        intermediate.right_stick_x = src.right_stick_x;
+        intermediate.right_stick_y = src.right_stick_y;
 
-        dst.dpad_up = src.dpad_up;
-        dst.dpad_down = src.dpad_down;
-        dst.dpad_left = src.dpad_left;
-        dst.dpad_right = src.dpad_right;
+        intermediate.dpad_up = src.dpad_up;
+        intermediate.dpad_down = src.dpad_down;
+        intermediate.dpad_left = src.dpad_left;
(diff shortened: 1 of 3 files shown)


rbrott:
Thanks for the detailed write-up and testing! The reasoning and patch itself seem good to me.

GraysonTinker:
Thanks! To clarify, the additional changes I suggested in regards to mapping guide on `STANDARD`, combining the two cases, and removing one or both of the additional checks didn't make it into this PR. Would you like me to make a new one with some or all of those changes?

PR #189, opened 2025-08-19
Robot Configuration Editor

ericwooshem:
This PR is an extension of #171 that allows the user to edit, create, and delete configurations through XML or GUI. This is a full implementation for #106.

Still working on further testing (and maybe there's hardware devices I missed).
Probably still going to change the GUI slightly, but here's the current state:

image
image

(window zoomed out to fit on one screen)


ericwooshem:
> Kudos to you for making it happen!

Thanks ❤️...  But I'm sure there's still tons of stuff wrong with it...

> ~I guess I'll have to find another project to wrap my FTC career with...~

😭 Sorry oop-

> One question I had that isn't immediately apparent from your screenshots or the code: how will the editor act when trying to modify/delete a readonly config?

Oh that's a really good point, I'll go try that.

ericwooshem:
> how will the editor act when trying to modify/delete a readonly config?

It allows the user to edit and save, but the file will get overwritten when re-uploading TeamCode. I think it would be better if didn't allow editing, I'll look into that.

ericwooshem:
> Maybe allow editing, but force the user to save it under a different name?

Oh yeah that makes more sense. Good idea!

ericwooshem:
Demo videos:

Creating and Saving:
https://github.com/user-attachments/assets/bc1526d7-4f9f-4a0e-9739-841ec57b3cba

Deleting a Config:
https://github.com/user-attachments/assets/5cb2350a-c50d-4fbf-a751-e3d1f0a32c5d

Read Only Config:
https://github.com/user-attachments/assets/ee35f514-1e5d-44da-bddb-2be0d8791bb8

Config Validation:
https://github.com/user-attachments/assets/40d6018c-ca6a-4f70-a167-50451c5e6dac

DashboardCore/build.gradle
```diff
@@ -21,6 +21,8 @@ dependencies {
     testImplementation "org.junit.jupiter:junit-jupiter:5.9.1"
 
     testImplementation "org.nanohttpd:nanohttpd-websocket:2.3.1"
+
+    implementation 'xmlpull:xmlpull:1.1.3.1'
```
rbrott: nit: double quotes here

DashboardCore/src/main/java/com/acmerobotics/dashboard/Mutex.java
```diff
@@ -24,4 +24,4 @@ public synchronized <V> V with(Fun<T, V> f) {
     public synchronized void with(UnitFun<T> f) {
         f.apply(obj);
     }
-}
+}
```
rbrott: nit: add back newline

rbrott:
Nice this looks good. I'm comfortable merging this unless you want to do more testing.

GraysonTinker:
If I could slip in one final suggestion, which you are of course welcome to decline, it would be to change the "Set" button to something like "Activate". I'm not quite happy with my original wording anymore, and this would be a bit clearer. Just a thought :)

PR #191, opened 2025-08-28
Deploy size optimizations: Significantly reduce the size of field images, remove legacy font formats

qwertychouskie:
This should cut off ~1MB from the APK size of the FTC SDK, reducing deploy/install times.

The Power Play field is much harder to compress without significant visual artifacts due to the gradient, and it's probably safe to remove.  The Into The Deep field compressed very well with basically no noticeable quality difference.

In the future, field images should probably by lossy .webp images like the Centerstage image is.


qwertychouskie:
I pushed another commit removing legacy font formats, which reduces the font directory size from ~578KB to ~66KB.

Unless we really need to target super ancient browsers like Internet Explorer, just shipping the woff2 format is sufficient.  (See https://github.com/FIRST-Tech-Challenge/FtcRobotController/issues/797#issuecomment-3231960166 for details.)

This brings us to ~1.5MB deploy size savings.

rbrott:
Thanks!

PR #192, opened 2025-09-03
Persist Graph Variable Selections Between Op Mode Runs

mxtmx:
This PR adds persistence for graph variable selections and window settings in the GraphView component. Previously, users had to re-select their telemetry variables every time they started a new op mode, which was tedious during iterative robot development.

### DIFF
--- a/client/src/components/views/GraphView/GraphView.tsx
+++ b/client/src/components/views/GraphView/GraphView.tsx
@@ -24,6 +24,7 @@ import { OpModeStatus } from '@/enums/OpModeStatus';
 import { colors, ThemeConsumer } from '@/hooks/useTheme';
 import { DEFAULT_OPTIONS } from './Graph';
 import { validateInt, ValResult } from '@/components/inputs/validation';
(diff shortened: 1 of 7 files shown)


rbrott:
This is a great feature! My main concern is that storing settings to local storage is a little weird with multiple graphs in a layout. What do you think about persisting keys between op mode runs but not between layout changes/reloads/app deploys and getting rid of local storage?

client/src/components/views/GraphView/GraphView.tsx
```diff
@@ -106,7 +110,7 @@ class GraphView extends Component<GraphViewProps, GraphViewState> {
 
     this.setState((state) => {
       if (this.props.telemetry.length === 0) {
-        return { availableKeys: [], selectedKeys: [] };
+        return { availableKeys: [], selectedKeys: state.selectedKeys };
```
rbrott, line 113: This seems like the only operative change.

I think it still makes sense to reset `selectedKeys` in this scenario when not graphing/at the selection screen. Otherwise there should be a mechanism for keeping the available keys alive (but then when would they ever be reset?).

PR #193, opened 2025-09-03
Fix All Linting Issues

mxtmx:
This PR fixes all the current linting issues:

1. **ConfigurableLayout.tsx**: Removed unused `clsx` import
2. **Graph.ts**:
   - Fixed TypeScript `any` types by using proper type assertions with `unknown`
   - Replaced non-null assertion (`!`) with proper null checking and error handling
3. **TelemetryView.tsx**: Fixed React Hook dependencies by using functional state updates instead of accessing current state values
4. **hardwareconfig files**: Fixed relative import issues by:
   - Adding `HardwareConfigState` to the centralized exports in index.ts
   - Updating imports to use the centralized `@/store/types` path instead of relative imports


rbrott:
Great, thank you!

PR #194, opened 2025-09-08
Add DECODE field image

bubner:
Courtesy to Team Juice 16236


bubner:
> What do you think about using the webp version? It looks to be a fair bit smaller than the png:
> 
> ```
> % du -h ~/Downloads/Juice-DECODE-Light.webp
>  68K    /Users/ryan/Downloads/Juice-DECODE-Light.webp
> ```

Good idea. I'll update it now

PR #196, opened 2025-09-12
add a struct interface + support to send telemetry packets using binary data encoded in struct format

zachwaffle4:
The struct format follows [wpilib's struct format](https://github.com/wpilibsuite/allwpilib/blob/main/wpiutil/doc/struct.adoc).
To be used effectively, we would need to add struct implementations for most of RoadRunner's geometry classes (which I'd be willing to do, especially because they mostly have equivalents in WPIMath already). 

The point of this PR is mostly to allow for sending data using structs to [AdvantageScope](https://docs.advantagescope.org/), but I also added support to decode the structured data on the front end. On the frontend it just looks like somebody invoked the `toString` method of a Kotlin dataclass (obivously except using the struct schema instead the constructor arguments).

### DIFF
--- a/DashboardCore/build.gradle
+++ b/DashboardCore/build.gradle
@@ -1,6 +1,7 @@
 apply plugin: "java"
 apply plugin: "maven-publish"
 apply plugin: "checkstyle"
+apply plugin: "kotlin"
 
 java {
     // TODO: add javadoc jar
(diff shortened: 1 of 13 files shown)


rbrott:
What do you see as the value proposition for dash here? Is the hope to save communication bandwidth for dash or turn telemetry packets into a universal API for third-party FTC telemetry?

zachwaffle4:
> What do you see as the value proposition for dash here? Is the hope to save communication bandwidth for dash or turn telemetry packets into a universal API for third-party FTC telemetry?

More of the second one.

> I suppose this is what I don't understand in particular. Why should dash care/know about AdvantageScope?

AdvantageScope is currently able to accept telemetry packets from FTC Dashboard/its webserver; right now a telemetry packet is just a string key + string value, but this PR also allows packets to include a struct schema and a list of byte arrays encoded according to the schema. AdvantageScope is able to parse that and use published pose data to draw on the field, for example.

PR #197, opened 2025-09-16
move log stuff here

zachwaffle4:
i also made it less ugly !


zachwaffle4:
> (also, my Kotlin RR log parser doesn't support logs v1 or arrays)
mine does :>

PR #198, opened 2025-09-22
add support for kotlin custom config variables

zachwaffle4:
so now u do 
```kt
object Something {
  var configVar by configurable("Something", "configVar", 0.0)
}
``` and it works yay wahoo

this uses reflection stuff like `ReflectionConfig` but also different somewhat so I had to write the entire thing again yayyyyyyyyyy


zachwaffle4:
i do not want to write annotation processors

BeepBot99:
That is not necessary, you can do it at runtime using Kotlin reflection.

Vlad1-1:
Just an opinion, I find this API much more native to kotlin.

However seeing kotlin properties from objects without @JvmField or @JvmStatic can be done from java by looking for the getProperty (or isProperty for Boolean properties) and setProperty functions in java reflection.

Also in my team's code we just use @file:Config at the top of the Constants file along with @JvmField on the properties to get top-level properties shown in the config menu. (it's shown as a ConstantsKt class generated by kotlin in bytecode)

PR #219, opened 2026-07-28
Add Graph CSV Export

Kellan-hub:
- Records graphed telemetry in the background while a run is graphed (only the checked keys, from when graphing starts to when it stops)
- Commits a CSV to a session-wide export list when graphing stops — manually or because the OpMode ends — one entry per run

- Adds a new Graph Export View, registered next to Recorder View in the view picker, with a dropdown + download button (same UI pattern as Recorder View) to download any export from the session
- Supports multiple graphed variables at once, each becoming its own CSV column
- Export list is in-memory only and clears on reload — never pruned automatically

Changes:
- GraphView now commits a recording on stop instead of discarding it
- New graphExports Redux slice
- New GraphExportView component
- Shared downloadBlob helper extracted from LoggingView (no behavior change there)
- View registered in ConfigurableView / ConfigurableLayout / ViewPicker

Testing:
- Manually verified against the mock server with single- and multi-variable OpModes
- Confirmed live multi-line graphing, both stop triggers, separate entries per run, correct CSV output, and list reset on reload

https://github.com/user-attachments/assets/9f72c0d6-6649-44e5-bfeb-8d9fe813aa2c

https://github.com/user-attachments/assets/1410bd5c-e865-449b-b26b-b9351ba7fcd9


rbrott:
What's the motivation for using this over the existing logging view?

Kellan-hub:
Ah I didn't think of logging view while creating this feature. As logging view accomplishes the same thing, I'll go ahead and close this PR

PR #220, opened 2026-07-29
Add opt-in battery current monitoring

Kellan-hub:
Adds a total current readout to the dashboard header, next to ping and voltage, behind a toggle. Off by default.

It's opt-in because it isn't free: LynxModule.getCurrent() blocks on a LynxGetADCCommand that bulk caching doesn't cover, so it's an extra round trip per hub per status poll. Teams who aren't looking at the number shouldn't pay for it.

Protocol: new SET_CURRENT_ENABLED message (client → robot); RobotStatus gains batteryCurrent in amps, -1.0 when off — same sentinel convention as batteryVoltage.

Robot: FtcDashboard holds a volatile boolean, and getRobotStatus() sums getCurrent() across every LynxModule. Voltage takes a Math.max in the same loop because the hubs share a rail; current sums because each hub draws its own load. The handler replies with a fresh status so the reading appears on toggle, not on the next tick.

Client: bolt button in the header with a pressed state; the readout appends  / x.xxA only while enabled. The reducer clears the stale value on toggle, and the socket middleware re-sends the flag on reconnect so the field doesn't go dead after a drop.

https://github.com/user-attachments/assets/795734f2-1d54-4acd-83cd-0e47b807feb5

### DIFF
--- a/.claude/settings.local.json.tmp.18512.6ad9c246a57b
+++ b/.claude/settings.local.json.tmp.18512.6ad9c246a57b
@@ -1,0 +1,24 @@
+{
+  "permissions": {
+    "allow": [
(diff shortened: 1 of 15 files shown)


rbrott:
Thanks for the contribution! I see some utility in putting this next to ping/voltage though I'm wary about the cost. It seems better for the user to decide how often to poll the current in their op mode and to put that in telemetry instead.

PR #223, opened 2026-08-05
Add Config dropdown on Top Bar

AlaniRongu:
- Add a hardware configuration selector dropdown to the dashboard topbar, letting users pick from available saved configurations (sorted alphabetically, with read-only configs labeled)
- Add an "Activate" button that dispatches the selected hardware config to the robot, disabled while an OpMode is running or when no unsaved change is selected
- Show an unsaved-change indicator (*) next to the selector when the selected config differs from the currently active one
- Show a confirmation modal ("Config Activated!") after activating a config, dismissible via the OK button, Enter, or Escape
- Track OpMode running state (via activeOpMode/activeOpModeStatus) to prevent activating a config mid-run

https://github.com/user-attachments/assets/4b86f9b4-4a69-48d3-9b71-eaea424fa2fc

### DIFF
--- a/client/src/components/Dashboard/Dashboard.tsx
+++ b/client/src/components/Dashboard/Dashboard.tsx
@@ -1,9 +1,12 @@
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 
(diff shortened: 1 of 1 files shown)


rbrott:
Thanks for the patch! I want to keep the top bar minimal and changing hardware config should be pretty uncommon so I don't plan on merging this.

